PR #1, opened 2015-05-11
[wip] Add script to generate configuration for tests, debug, perf problems...

ddurieux:


Seb-Solon:
Jinja should be user for generation.

See https://github.com/savoirfairelinux/monitoring-tools/blob/master/tools/new-plugin.py for jinja use
and 
https://github.com/savoirfairelinux/monitoring-tools/blob/master/tools/templates/python/tests/test_check_foo.py
and 
https://github.com/savoirfairelinux/monitoring-tools/blob/master/tools/templates/python/shinkenplugins/plugins/foo/__init__.py

for templates. 

basically you use {{ var_name }} in the template and it will replace it.

naparuba:
http://search.cpan.org/~nierlein/Monitoring-Generator-TestConfig-0.44/lib/Monitoring/Generator/TestConfig.pm ?

ddurieux:
It's better to have it in python, hate to have too multiple langages scripts in the repository ;)

PR #3, opened 2015-05-13
Add setproctitle module in required

ddurieux:


shinken/misc/common.py
```diff
@@ -80,7 +80,4 @@
 
 }
 
-try:
-    from setproctitle import setproctitle
-except ImportError as err:
-    setproctitle = lambda s: None
```
Seb-Solon: I would raise  a proper Error exception instead of letting it crash. So that there is no "big" behavior difference

gst:
This PR should only have  `+setproctitle` in the requirements file then ;)

The others "blank" changes (space / TAB) should be in a separate PR if you want to clean that kind of spaces / tabs ..

otherwise +1 ;)

Also : yes, I think we can keep the `try: import setproctitle except ..` for now if not for always. It can effectively be better to keep it in case something is wrong with the installation of this library and that it can't be imported.
**But** : you could add a `warnings.warn(..)` in the except clause to have a trace of the actual import error.

gst:
@ddurieux : ping ?

Seb-Solon:
Commit have to be squashed IMO

gst:
yep, squash that, rebase it on develop, and it'll be merged ;)

gst:
this can be closed per #67

PR #9, opened 2015-05-19
Better str and repr for objects

gst:
So to have better view/repr of objects when in debug sessions.

Main thing done is : 
don't anymore return `str(self.__dict__)` (or `str(self.items)` for Items classes) for the \_\_str\_\_ or \_\_repr\_\_ methods of Shinken objects. 
That was more overkill than useful (and has even caused some problem when using that "feature" in debug sessions).

### DIFF
--- a/shinken/objects/command.py
+++ b/shinken/objects/command.py
@@ -96,9 +96,6 @@ class Command(Item):
     def get_name(self):
         return self.command_name
 
-    def __str__(self):
-        return str(self.__dict__)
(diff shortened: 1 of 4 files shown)


gst:
We could eventually squash all that together ..   wdyt ?

*but* keep the last commit on its own 'cause it's better to keep it separate as while being related (it's within an \_\_str\_\_ method) the reason of it is another issue.

PR #11, opened 2015-05-19
Better str and repr for objects

gst:
So to have better view/repr of objects when in debug sessions.

Main thing done is : 
don't anymore return str(self.\_\_dict\_\_) (or str(self.items) for Items classes) for the \_\_str\_\_ or \_\_repr\_\_ methods of objects. 
That was more overkill than useful (and has even caused some problem when using that "feature" in debug sessions).

### DIFF
--- a/shinken/objects/item.py
+++ b/shinken/objects/item.py
@@ -212,6 +212,8 @@ Like temporary attributes such as "imported_from", etc.. """
         cls_name = self.__class__.__name__
         return '<%s "name"=%r />' % (cls_name, self.get_name())
 
+    __repr__ = __str__
+
     def is_tpl(self):
         """ Return if the elements is a template """
         return not getattr(self, "register", True)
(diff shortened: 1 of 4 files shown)


shinken/objects/item.py
```diff
@@ -205,8 +205,14 @@ def clean(self):
...
 
+    def get_name(self):
+        return getattr(self, 'name', "unknown")
+
     def __str__(self):
-        return str(self.__dict__) + '\n'
+        cls_name = self.__class__.__name__
+        return '<%s "name"=%r />' % (cls_name, self.get_name())
+
+    __repr__ = __str__
```
ddurieux: Why this line is alone (not in the function) ? seems bad

gst: this simply define `__repr__` as an alias for `__str__`, without having to do something like:
```
def __repr__(self):
    return self.__str__()  # or better : return str(self)
```

PR #14, opened 2015-05-19
Cleaned and removed compute_hash() related.

gst:
Not anymore used since some time. 
but was still unnecessarily called / computed grrrrr ..


gst:
I'll check about the failed travis build

gst:
temporary failed travis build was due to port not free or things like that.

So it's ok: up to your merge dudes ! @ddurieux @Seb-Solon   :p

PR #15, opened 2015-05-19
Enh simplify commandcall serialization

gst:
This is to be seen / think about and eventually tested in a real setup..

3rd commit is optional.


gst:
Ho, you ~~made~~ merged it directly !

Well, we'll see if that doesn't trigger any side problem (which I'm not sure about) sooner than later thus ;)

gst:
Not 100% sure but I think I've encountered a side effect..  have to check.

PR #16, opened 2015-05-19
Fix bad is comparison

gst:
Fix a bad "is" comparison

\+ reviewed and slightly cleaned in satellite code : 

- do_manage_returns
- do_pynag_con_init
- manage_returns
- get_return_for_passive

no more, no less.


gst:
Temporary port not free as well.. damn it occurs more often now it seems.. looking into that..

PR #17, opened 2015-05-19
Enh len of list comprehension to len of generator

gst:
Many pythonic changes which should result in many (small but still) performance improvement changes..

I measured most of them but some aren't that obvious.. concerning the perf improvement.  
if one want we can talk about it..

### DIFF
--- a/alignak/daemons/arbiterdaemon.py
+++ b/alignak/daemons/arbiterdaemon.py
@@ -556,7 +556,7 @@ class Arbiter(Daemon):
             stats['nb_' + t] = nb
             logger.info("Got %s for %s", nb, t)
 
-        max_srv_by_host = max([len(h.services) for h in self.conf.hosts])
+        max_srv_by_host = max(len(h.services) for h in self.conf.hosts)
         logger.info("Max srv by host %s", max_srv_by_host)
         stats['max_srv_by_host'] = max_srv_by_host
 
--- a/alignak/daemons/arbiterdaemon.py
+++ b/alignak/daemons/arbiterdaemon.py
@@ -552,7 +552,7 @@ class Arbiter(Daemon):
                  'realms']
         for t in types:
             lst = getattr(self.conf, t)
-            nb = len([i for i in lst])
+            nb = len(lst)
             stats['nb_' + t] = nb
             logger.info("Got %s for %s", nb, t)
 
--- a/alignak/daemons/schedulerdaemon.py
+++ b/alignak/daemons/schedulerdaemon.py
@@ -157,22 +157,17 @@ class IStats(Interface):
 '''
     def get_raw_stats(self):
         sched = self.app.sched
-        nb_scheduled = nb_inpoller = nb_zombies = 0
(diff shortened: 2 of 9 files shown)


ddurieux:
Add test for cover these functions please to be sure these functions are not broken

gst:
> Add test for cover these functions please to be sure these functions are not broken

all these changes are really "*easy*" changes and according to the travis build they aren't broken ;)

gst:
what surprised me is that : 

`len([ x for x in sequence if filter(x) ])`  forms aren't necessarily always slower than the more pythonic form `sum(1 for x in sequence if filter(x))`  (which doesn't creates unnecessarily a temp list before doing it's len()).

but the difference is very little when the former is quicker than the later.

gst:
> what surprised me is that :

ah ! it depends how the sum() argument is constructed !

```
In [7]: timeit.timeit("len([x for x in lst if filt(x)])", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=100)
Out[7]: 0.13385295867919922

In [8]:

In [8]: timeit.timeit("sum(filter(filt, lst))", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=100)
Out[8]: 0.1352519989013672

In [9]:

In [9]: timeit.timeit("sum(filter(filt, lst))", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=100)
Out[9]: 0.1256258487701416

In [10]: timeit.timeit("sum(filter(filt, lst))", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=100)
Out[10]: 0.1333022117614746

In [11]: timeit.timeit("sum(filter(filt, lst))", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=1000)
Out[11]: 0.9786028861999512

In [12]: timeit.timeit("len([x for x in lst if filt(x)])", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=1000)
Out[12]: 1.2255499362945557

In [13]: timeit.timeit("len([x for x in lst if filt(x)])", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=1000)
Out[13]: 1.2320811748504639

In [14]: timeit.timeit("sum(filter(filt, lst))", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=1000)
Out[14]: 0.9566860198974609

In [15]: timeit.timeit("sum(filter(filt, lst))", "lst = list(range(10000)) ; filt = lambda v: v%13!=0", number=1000)
Out[15]: 1.0634419918060303
```

So I'll make another pass and see if I can get that.

gst:
so sum() with an "if-generator" `sum(x for x in lst if y(x))` is a little slower than sum with a filter() applied to the generator : `sum(filter(lst, y))`  and this one is even more pythonic than the previous.. So I'll made an update.

gst:
rebased on develop..

Seb-Solon:
"little DRY" is IMO not a valid commit message.

PR #19, opened 2015-05-20
update shebang to '#!/usr/bin/env python'

ddurieux:


aviau:
The test is overkill. 

If you decide to keep it,  create a test that will be executed by nosetests.

gst:
yeah but better overkill than have nothing at all.

I'm gonna merge. we can simplify this test later if really necessary.

PR #26, opened 2015-05-22
Enh/Fix: make modules still compatible with us  [WIP]

gst:
Used the most simple trick : load all alignak modules and use them as "alias" in sys.modules.

An import hook could also have been used I'm pretty sure, which could have been better (I gave it a try but had slight problems so I've fallback on the above possibility).

### DIFF
--- a/alignak/__init__.py
+++ b/alignak/__init__.py
@@ -42,3 +42,7 @@
 #
 #  You should have received a copy of the GNU Affero General Public License
 #  along with Shinken.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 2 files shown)


gst:
NB: with this I've my first Alignak instance running correctly with few modules OK.

gst:
Ah !

I could come to solution with an import hook which now seems ok..   maybe it's better (or not ?).. gonna test it a bit more and if still ok then I'll open another PR with it..

gst:
closed in favor of  #27.

PR #27, opened 2015-05-22
Enh/Fix: make modules still compatible with us

gst:
This one should be better than https://github.com/Alignak-monitoring/alignak/pull/26

because this one is dynamic (and so does not preload/import every possible module).

### DIFF
--- a/alignak/shinken_import_hook.py
+++ b/alignak/shinken_import_hook.py
@@ -1,0 +1,27 @@
+
+import importlib
+import sys
+
+
+class Finder(object):
+
+    def find_module(self, fullname, path=None):
+        if fullname == 'shinken' or fullname.startswith('shinken.'):
+            return self
+
+    def load_module(self, name):
+        mod = sys.modules.get(name)
+        if mod:
+            return mod
+        alignak_name = 'alignak%s' % name[7:]
+        mod = sys.modules.get(alignak_name)
+        if mod:
+            sys.modules[name] = mod
+            return mod
+        mod = importlib.import_module(alignak_name)
+        sys.modules[name] = mod
+        return mod
+
+
+finder = Finder()
+sys.meta_path.append(finder)
--- a/alignak/__init__.py
+++ b/alignak/__init__.py
@@ -42,3 +42,6 @@
 #
 #  You should have received a copy of the GNU Affero General Public License
 #  along with Shinken.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from . import shinken_import_hook


gst:
@Seb-Solon  ..

which one do you prefer ?

Seb-Solon:
This one is better IMO

alignak/shinken_import_hook.py
```diff
@@ -0,0 +1,27 @@
+
+import importlib
+import sys
+
+
+class Finder(object):
+
+    def find_module(self, fullname, path=None):
+        if fullname == 'shinken' or fullname.startswith('shinken.'):
+            return self
```
ddurieux: module with shinken ?

ddurieux:
Not really why you do this?
You have problem on load alignak modules?

aviau:
We don't want to make our modules incompatible with shinken.

Seb-Solon:
shinken module usually does from shinken import blabla. That's the issue. This comes from the fact you change the folder and everything to alignak when renaming instead of only editing header. We have to adapt the code now to be able to load shinken modules.

ddurieux:
I understand that, but:

* For the moment shinken and alignak have same structure, in future no
* Alignak will be in python 2.7 and 3.4 (currently in progress), so all shinken module will be compatible
* Modules for alignak will be installed only with 'pip install', not the case with shinken 
* Modules fro broker may be all in external (internal is deleted, work in progress), perhaps can do same with other modules
* We want too some module dropped because have too many module is not a good idea. You can see all retntion modules of shinken and more than half not work

Seb-Solon:
I don't get it. This PR was a technical fix because otherwiase shinken module can't work with alignak (and vice versa) 
The fact all module have a import shinken / import alignak force us to hack this.
Otherwise, we have to right now fork all module we need and replace those import. 
This PR is to keep everything working for now (because @aviau for example may need it). Of course we should remove this later. 

Point 2 3 and 4 are not related to this PR IMO.

ddurieux:
You don't think it's like open the Pandora's box ?

aviau:
> You don't think it's like open the Pandora's box ?

Is there any downsides to making modules compatible?

ddurieux:
it will not be complex (mean problems to maintain, specific errors, specific features) with time and divergence of code between alignak and shinken ?

aviau:
Do we plan on heavily changing the modules API?

If yes, are we doing it right now?

Until this happens, why not use the same modules?

gst:
> I understand that, but:
> ...

that's all right I think.

This PR is only for being able to run, without a change at them, shinken modules *up to / until* we'll change this API (if we need to, not impossible.).

gst:
So ok for merge ? who wants ?

ddurieux:
I'm ok to merge but before, can you add an enhance issue with these points: 

* ref to this PR to clean later
* add comment to delete shinken.io ref in code (because modules will be installer via pip install).

PR #29, opened 2015-05-22
Regenerate licence headers in python file with good contributors

Seb-Solon:
Hi,

Most of file had a copy pasted header from Shinken, I generated the Shinken part of the new header with the script in contrib and json.

It is based on the git log of files so everybody is in the file he contributed to.

The is only a small cleaning left to do : detelete "TODOLETE" lines

I think we have to do it for shell script also but it can be done manually.

You can start to review that because it is a big one.


ddurieux:
Perhaps create PR only when you have cleaned the TODOLETE ;)

Seb-Solon:
No because I wanted someone to start reviewing that as it is a big diff.

ddurieux:
I prefer take time more but less times ;)

gst:
> > Perhaps create PR only when you have cleaned the TODOLETE ;)
> No because I wanted someone to start reviewing that as it is a big diff.

I agree with Seb : sometimes you want to make a PR to eventually propose some change but want to have advices or comments on it *before* going further..  (putting "[WIP]" in the PR title is a good habit in such case).

gst:
code seems good to me..

but.. but should it be really part of the project ??  it's only a one-time use tool afaik and I'm not sure it's necessary to have it here.. Am I wrong ?

Seb-Solon:
I agree, I did not know how to push that and need to version it.

gst:
well, my points/questions are really more "details" or not that important..

So if you want you can ignore them ;)

ddurieux:
For the future commit in alignak, only put your name in AUTHORS file. the alignak header link to this file

Seb-Solon:
Yep, it will be less painful :)

PR #33, opened 2015-05-27
Fix setup.py

gst:
So I've rebased the original branch done by @titilambert  (on our develop one).

I had then to continue some needed adaptations/fixes.

something I'm not sure about (don't know pbr more than that) :  doing "python setup.py install" doesn't install the requirements.. I had to use "pip install ."  (for setup_test.sh+.travis.yml)..
Same kind of prob for the test-requirements : had to keep/handle that in setup_test.sh. (while I'm rather sure it should be doable from/with the setup.py+setup.cfg themselves).

otherwise some simple path fixes (I have restored "daemon" scripts into alignak/bin so that they are importable by the "bin" scripts (installed by pbr (entry_points -> console_scripts)).

also removed the "debian-ugly" from setup.py.. not sure there are not other to be removed/cleaned (like the one in install_hooks.py )

waiting comments..

If that's good to you :  I'm hesitating to eventually squash all commits together before merge.. could be more clean for this PR..


gst:
@titilambert 

can you take over it ?

gst:
FYI:  I've implemented a better version "handling" within https://github.com/gst/alignak/commit/4ab4dce96e7dd268eb6eefc6ab8e5902d7cbeb2a

in the process of cleaning at most everything I think we should have that too ; either directly in this PR or in a separate one..

Seb-Solon:
Also fix naparuba/shinken#1727 :)

Seb-Solon:
Related to naparuba/shinken#1561

PR #34, opened 2015-05-27
[wip] Clean: dropped the stats module and all its usage

gst:


Seb-Solon:
I disagree. I think we should drop only the part that send it to an external website. The fact we want to collect stats is good. Maybe it's a module job but still the data collect is necessary.

gst:
okay, will see that.

ddurieux:
Yes performances metrics can be important. I don't think it's possible to use a module for that.

mohierf:
I agree with @Seb-Solon and @ddurieux ... collecting metrics from Alignak core activity is important. Those metrics should be sent to an external module which will forward to Graphite or Statsd ?

ddurieux:
We have removed specific statsd to external website, now have only statsd functions

PR #38, opened 2015-05-27
Clean: reuse bottlecore to implement bottlewebui

gst:
which had only that as difference from bottlecore..

> 2,818 deletions,
> 6 additions,


gst:
the next (easy) thing to do with this will be to remove the copy of bottlecore there is in alignak/webui ; and use the one from pip/package/..

this can, more cleanly, be done in a separate PR.

Issue #22, opened 2015-05-21
Remove the shebang where it shouldn't be


gst:
Only files aimed to be executed should have a shebang line.

any other ~~file~~ module (let it be part of package or not) should **not** have that shebang.

gst:
closing, this test is/would be, imho, simply overkill and doesn't bring any valuable info/result.

PR #39, opened 2015-05-27
clean: removed importlib from our source

gst:


alignak/shinken_import_hook.py
```diff
@@ -1,5 +1,5 @@
-
 # -*- coding: utf-8 -*-
```
Seb-Solon: Careful. I got the encoding line at line n3/4 previously and it was not working. I don't know if it's in the first line it is stil OK

gst: it's ok : https://www.python.org/dev/peps/pep-0263/ : 

```To define a source code encoding, a magic comment must
    be placed into the source files either as first or second
    line in the file```

second line is clearly for the case where you use a shebang .. so in this case (no shebang needed) the good/better choice is the first one ;)

ddurieux: Why no shebang needed?

gst: simply because it's not a *script*, it's a simple python *module*, as are 99(.9)% of the *.py* files  in the project.  And so should not be directly **executed** as a python script.
Even the daemons *main* modules (which are actually under `bin` : `arbiter-daemon.py` and the others..) (but that will be (when the setup PR will be done)) in *alignak/bin/* (or more precisely in `alignak.bin`) won't require the shebang, as they will be **imported** (not directly *run/executed*) by the console_entry scripts setup by the .. setup(pbr) ;)

*basically* : if there is no `if __name__ == __main__:` in a module then it is not a *script* (or should not be treated as one). And even when there such a construction it's not always necessary/required/desired to have the shebang, like when you want to execute a package (`python -m package(.sub..)`).

gst: that's also the reason of https://github.com/Alignak-monitoring/alignak/issues/22 about the clean of unnecessary shebangs in all the python modules..

ddurieux: Ok I understand now, thanks for the detail ;)

gst: you're welcome ;)

ddurieux:
importlib is only in python3, so we will have problem with python 2.7 no?

gst:
we have it in requirements.txt

gst:
it's only in python2.6 that's it's not available by default.

ddurieux:
Oh yes, great ;)

PR #40, opened 2015-06-01
Added coverage badge to README

gst:
TODO: make coverage build to succeed


aviau:
Fix coverage before you merge this,

gst:
actually I wanted to use the badge to see / know when the "build" will be ok.. but well..

gst:
what's quite boring is that coveralls.io is *very* slow at generating its results..

aviau:
You can go on the coveralls website. You dont need the badge to see if the build works.

gst:
yeah but coveralls.io, not only *very* slow, is also not that clear .. what I want to get is the badge status ok (whatever is the %) ; so my first merge of it before all ;)

PR #52, opened 2015-06-04
Enh: execute a pip freeze before nosetests

gst:
so to know what exact libs/packages versions are in use.

It **is** usefull for debug.


Seb-Solon:
I'm Ok with that : we have a list  with version explicitely instead of looking at the pip install -r logs

PR #54, opened 2015-06-05
Enh: handle version info as in PEP-0440 + PEP-0396

gst:
Basically:

1) store our version number in alignak.version.VERSION
   -> so it can be executed by setup to get it.

2) but, as requested by PEP-0396 : make it available through alignak.\_\_version\_\_

3) make deprecated (but still usable) the old one we used (alignak.bin.VERSION)

https://www.python.org/dev/peps/pep-0440
https://www.python.org/dev/peps/pep-0396


gst:
aaahh, I just spotted one place in alignak were the old version path was still used..  okay will make another PR to update it to the new more correct one..

PR #62, opened 2015-06-08
Fix coverage broken after setup merge

gst:


.travis.yml
```diff
@@ -13,7 +13,11 @@ install:
 script:
   - cd test
   - pip freeze  # so to help eventual debug: know what exact versions are in use can be rather useful.
-  - nosetests -xv --process-restartworker --processes=1 --process-timeout=999999999  --with-cov --cov=alignak
+  - which nosetests
+  - nosetests -V
```
Seb-Solon: Do you want to keep those lines? IMO we sould put that in a script because it is basically 5 lines giving only info about the system

Seb-Solon:
Squash Squash

.travis.yml
gst: could be put in setup_test.sh but what's wrong with having it in travis ?  it's really with travis that I wanted to be sure of the versions used.. so my add in .travis file :)

gst:
not squashed, it's really better to have few little commits than one bigger, when the little commits are not "do-undo-redo-etc.." of the same thing/change(s) but effectively different things/changes.

gst:
what I'm not sure is to keep/postpone the last commit : "try multiprocess for nosetests" because I know there are times when this apparently triggers some timerace problems.. though not always..

gst:
> what I'm not sure is to keep/postpone the last commit : "try multiprocess for nosetests" because I know there are times when this apparently triggers some timerace problems.. though not always..

but on the other side it really speed up the travis build, up to 2x factor (travis build runners seem to always have 2 cpu available), so if the "timerace" problems I've already encountered doesn't happen more frequently than say once for 4 then it's still a good win .. and actually I'm pretty sure it's the case..

.travis.yml
Seb-Solon: Well readibility purpose IMO. Add some which travis and so is As ugly as letting print statement in python or letty -x in bash.

Seb-Solon:
>it's really better to have few little commits than one bigger, when the little commits are not "do-undo-redo-etc..

Last 3 commits (943e100 1d5fe3e 2e23caa) are "drafts" because you actually do and undo stuff

>Fix coverage broken after setup merge

What is broken? 
Is removing removing alignak import file part of fix / the cause of "break" ?

>I know there are times when this apparently triggers some timerace problems

What kind of issue?

gst:
>> I know there are times when this apparently triggers some timerace problems
> What kind of issue?

the kind you've encountered where you have to restart a blocked travis build :p

gst:
squashed a bit and removed the "which coverage and nosetests" + their versions.

Seb-Solon:
Here we go!

PR #63, opened 2015-06-08
Clean daemons & http shutdown

gst:
Basically: join the main http thread (and in it join all the http client threads) when stopping an alignak daemon, so that no http thread is left running while everything is closed (which can, and have already, be harmful) .

\+ some cleaning 
\+ some better logs 
\+ few fixes related to http.

### DIFF
--- a/alignak/basemodule.py
+++ b/alignak/basemodule.py
@@ -315,7 +315,7 @@ class BaseModule(object):
 
         # TODO: fix this hack:
         if alignak.http_daemon.daemon_inst:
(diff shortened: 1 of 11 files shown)


gst:
don't merge directly... have to verify some thing..

gst:
ok that's a first quick clean / fix ; I think it's preferable to merge it as it.. 

There will be anyway others passes and (big) clean/changes/rewrite on the http_daemon module so let's go..

PR #73, opened 2015-06-12
Add: Test for setup.py. There are some hacks for Travis

Seb-Solon:


ddurieux:
It's possible to have script in sh instead bash, because I don't use / want to install bash in my FreeBSD?

Seb-Solon:
Well can we do it in another PR ? There are some bash substitution I have to find how to remove.

gst:
I think I'm ok for the merge.. 
@Seb-Solon  you had not some things to finish / enhance ? 
if not then I'll merge..

Seb-Solon:
If @ddurieux is ok to change to sh in another Pr yeah we can merge :)

PR #74, opened 2015-06-12
Enh/Fix: replace a possible infinite while loop by a for loop.

gst:
create_and_launch_worker() does NOT always add a queue to self.q_by_mod,
so the `while len(self.q_by_mod[mod]) < self.min_workers:` could loop forever.

Use a for loop and that's it.

Check create_and_launch_worker() to see that it has "premature" return..

### DIFF
--- a/alignak/satellite.py
+++ b/alignak/satellite.py
@@ -494,7 +494,8 @@ class Satellite(BaseSatellite):
 
     # Create and launch a new worker, and put it into self.workers
     # It can be mortal or not
-    def create_and_launch_worker(self, module_name='fork', mortal=True):
+    def create_and_launch_worker(self, module_name='fork', mortal=True,
+                                 __warned=set()):
         # create the input queue of this worker
         try:
             if is_android:
@@ -523,6 +524,10 @@ class Satellite(BaseSatellite):
                         raise NotWorkerMod
                     target = module.work
             if target is None:
+                if module_name not in __warned:
+                    logger.warning("No target found for %s, NOT creating a worker for it..",
+                                   module_name)
+                    __warned.add(module_name)
                 return
         # We want to give to the Worker the name of the daemon (poller or reactionner)
         cls_name = self.__class__.__name__.lower()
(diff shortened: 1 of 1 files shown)


gst:
NB: not created a test case for this : basically because it would loop infinitely or else it would require to mock quite agressively the function create_and_launch_worker() which wouldn't have any sense left.

Seb-Solon:
I don't think having a use case integrated in nose test is a good idea. However I think posting here the way to reproduce could be helpgul to notice the diffrence if we want to manually try this fix.

gst:
> However I think posting here the way to reproduce could be helpgul to notice the diffrence if we want to manually try this fix.

don't get you..

Anyway : use "native_plugins" (https://github.com/shinken-monitoring/mod-native-plugins) with the poller and this infinite loop is triggered..

Seb-Solon:
Yes, that what I meant. If a pull requst is fixing a bug we should provide a way to reproduce it.

So here the setup would be : 
   * Install native plugin module
   * Use native plugin module in poller
   * Wait for the fail in logs

Right?

gst:
right :)

Seb-Solon:
Actually, I reviewing this again  (doing docstring) 

I dont understand the use of  __warned. It is a local variable only edited when it's None but then you directly return so that this variable is removed. I guess you did not want to warn several type for the same module but it won't work in this case.

Am i right @gst?

gst:
you are right for the reason of why I've used this (to not have several warns for the same `module_name` (otherwise they ~~would be~~ are logged again and again (and again..) at each turn of the scheduler main loop....)). 

But this function variable is not so *local*.. because it's part of the **function definition/signature** (*and have a default value assigned*, ofcourse/obviously).. and "so" (actually) it's only evaluated once (when the module of that function is actually imported the first time and that function definition is evaluated), so the actual `set()` assigned at that time is always the same one which is reused for each call to the function. quite common "caching" pattern (for a function):  http://docs.python-guide.org/en/latest/writing/gotchas/#when-the-gotcha-isn-t-a-gotcha

Seb-Solon:
Oh you right, that mutable stuff :) I will docstring it good because I feel like you are exploiting something :)

gst:
Always exploit the power of the language !!

Issue #60, opened 2015-06-08
fix / clean / rewrite the http_client module


gst:
Don't use anymore pycurl.

I propose to use the requests (http://docs.python-requests.org/) module.

gst:
linked to https://github.com/Alignak-monitoring/alignak/issues/53

ddurieux:
I'm not against this, but what we will win with python-requests instead curl?

gst:
requests is greatly more appropriate / easy / pythonic, 

and also it's usable/compatible with `pypy` while pycurl isn't ; which is the main blocker for us to try it (we could - quite very easily - obtain lot of performance wins)

http://stackoverflow.com/questions/15461995/python-requests-vs-pycurl-performance -> http://stackoverflow.com/a/15462207/2451329

so I would like to take the opportunity to completely clean the http_client module and rewrite it using requests.

gst:
ongoing..  GET & POST seems already good..

gst:
Seems good.. *seems* ..

gst:
has been merged !

PR #80, opened 2015-06-12
Enh/clean: replaced pycurl by requests in HTTPClient.

gst:
Almost complete rewrite of http_client module using requests instead of pycurl.

already quite simpler and, more important certainly, this will allow us to try pypy, very soon..

NB: the http_client module is not **specifically** tested as far as I know / see .. 
though, 45% of it is covered by some test : https://coveralls.io/builds/2802007/source?filename=alignak%2Fhttp_client.py

but I've run my alignak dev daemons with this commit and the conf / etc were correctly dispatched, etc..

### DIFF
--- a/alignak/http_client.py
+++ b/alignak/http_client.py
@@ -46,30 +46,20 @@
 #  along with Shinken.  If not, see <http://www.gnu.org/licenses/>.
 
 import cPickle
+import warnings
 import zlib
-import json
 
-# Pycurl part
-import pycurl
-pycurl.global_init(pycurl.GLOBAL_ALL)
-import urllib
-from StringIO import StringIO
+import requests
 
-from alignak import __version__
+
 from alignak.log import logger
-PYCURL_VERSION = pycurl.version_info()[1]
+
 
 class HTTPException(Exception):
     pass
 
 
 HTTPExceptions = (HTTPException,)
-
-class FileReader:
-    def __init__(self, fp):
-        self.fp = fp
-    def read_callback(self, size):
-        return self.fp.read(size)
 
 
 class HTTPClient(object):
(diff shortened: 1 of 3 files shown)


gst:
will fix https://github.com/Alignak-monitoring/alignak/issues/60

gst:
comment ?

Seb-Solon:
Well, you will tell me I'm bothering you with test but IMO this change is kinda risky because we don't have functional tests right now.

> the http_client module is not specifically tested as far as I know / see .. 

The coverage for your PR is 58% : https://coveralls.io/builds/2801982/source?filename=alignak%2Fhttp_client.py better then previously actually. 

I can see in both cases only the get seems to be covered. What about post and put? I don't know where the get is covered but that could be great to see if we can have the other called. What do you think?

gst:
effectively only the `get` is covered (by which test is something else, I'd guess by more than one probably)..

but I agree that the `post` and `put` which are defined (&used) *should* be as well be covered by some test(s) and even explicitely (and so the `get` as well)..  that was never done yet ;)

problem is that the state of the `HTTPDaemon` (the server part against which this http_client applies) doesn't necessary makes it easy.. though.. well in fact I started writting one new test file doing exactly this actually.. I was a bit after its fresh start and was at the point to think on how to more easily get to the `put` and `post` (and `get`) of an HTTPClient instance being usable in some easy configuration so in a TestCase instance..

give me some minutes and I'll check if I can get above very quickly advanced .. :p

gst:
in fact, IMHO, the "http_daemon" (and its related `http_thread` (attribute of Daemon) as well ; in fact http_thread & http_daemon objects should be merged into one single entity/object) **should** be completely rewritten as this http_client is now done..

also the fact that there is 2 possible http backends (which have some differences in how they run which leads to some distinct problems between them.. ho well..) is.. err.. don't know..
and otherwise the "quality" of that http_daemon file is not very good.. really.

but well.. I give a (fast) try on my previous comment first (we never know it'd be that easy.. (hmm)).

gst:
ok I could got something good and quite easy.. but had not time left to push it.. I'll try to do that tonight or later..

gst:
+ tests added, 

+ executed with pycurl old http_client version -> OK  (when there isn't some weird timing issues blocking the full test.. but I now know more about them and will be able to provide a fix).
+ tests with requests version are also ok (if no timing issue, otherwise : relaunch build/tests).

PR #97, opened 2015-06-17
Add alignak and alignak.modules namespaces

titilambert:
@gst @Seb-Solon @ddurieux  
What do you think about this ?
Which module can I use to test it ?

### DIFF
--- a/alignak/__init__.py
+++ b/alignak/__init__.py
@@ -48,3 +48,6 @@ from .version import VERSION as __version__
 
 
 from . import shinken_import_hook
+
+
+__import__('pkg_resources').declare_namespace(__name__)
--- a/alignak/modules/__init__.py
+++ b/alignak/modules/__init__.py
@@ -1,0 +1,1 @@
+__import__('pkg_resources').declare_namespace(__name__)
--- a/setup.py
+++ b/setup.py
@@ -26,5 +26,8 @@ os.environ['PBR_VERSION'] = VERSION
 setuptools.setup(
     setup_requires=['pbr'],
     version=VERSION,
+    packages=['alignak', 'alignak.modules'],
+    namespace_packages=['alignak', 'alignak.modules'],
     pbr=True,
 )
+


gst:
that's almost or already completely all what is needed I think (to be double-checked)..

but I'm thinking to it.. as you know : I've a **little** concern about the `alignak´ (top-)package's `\_\_init\_\_` where we require some actions to be done (even if it's nothing big, at all, but still)..

what if we would use a *dedicated* package name for this, something like "alignak_ns" (better name possible I'm pretty sure), then we would not have the problem.. we would have so `alignak_ns.modules`, `alignak_ns.plugins`,  etc ....

hmm.. unfortunate that the namespace feature isn't possible in a sub-package name (that is 'alignak.modules' without having the parent (that is 'alignak') to also be a namespace.. otherwise I'd be +2 for merge..
Now I'm "only" +1 for this merge ;)

gst:
> but I'm thinking to it.. as you know : I've a little concern about the alignak (top-)package's __init__ where we require some actions to be done (even if it's nothing big, at all, but still)..

to be complete about the concern I have : in the namespace package feature : all and every `__init__` modules from the root package (`alignak`) to the every "namespaced" package (`alignak.modules` in this case) isn't guaranteed to be run when the package is imported..  which is my concern.. now have to "double-triple" check this in all possible configuration (actually py26 + py27)..

titilambert:
See refs: https://docs.python.org/2/tutorial/classes.html

gst:
quoted from https://www.python.org/dev/peps/pep-0420/#id22 : 

"""
 The recommended way of using it is to put:

from pkgutil import extend_path
\_\_path\_\_ = extend_path(\_\_path\_\_, \_\_name\_\_)
in the package's __init__.py . Every distribution needs to provide the same contents in its __init__.py , so that extend_path is invoked independent of which portion of the package gets imported first. **As a consequence, the package's \_\_init\_\_.py cannot practically define any names** as it depends on the order of the package fragments on sys.path to determine which portion is imported first.
"""

titilambert:
What do you think about these namespaces:
alignak
alignak.core (or alignak.common): alignak
alignak.modules: all modules
alignak.plugins: all plugins

ddurieux:
What the difference between modules and plugins?

titilambert:
Plugins could be used by something like this: https://github.com/shinken-monitoring/mod-native-plugins

ddurieux:
It's not very clear for me sorry. Not really understand 'plugin' term

Seb-Solon:
@ddurieux plugin = monitoring-plugin as we know. It not a big deal to put those in a specific namespace, you still can run them in a "standalone" way

Is it more clear?

ddurieux:
ok, plugin is not enough explicit.
It's better with name like 'commandcheck' (not sure this name is the better but we better understand the goal and what for it is used)

gst:
2 cents begin: concerning my little trouble : it should not be a blocker; the version number which is available in `alignak` package, isn't per se necessary/required to be available exactly there.. that's better but that's not an hard requirement. And for the shinken import hook it can also be put in another sub-module of alignak, (as far as it's imported when a daemon is imported (´alignak.bin`).
2 cents finished.

PR #106, opened 2015-06-20
Enh: Move alignak/clients/* to contrib/clients

Seb-Solon:
This should not be installed on a system, those file are not useful for alignak to work


ddurieux:
Seems it's for livestatus, so remove it completly?

Seb-Solon:
I was planning to decide that when we will deal withe the whole contrib file ^^

ddurieux:
oki

PR #109, opened 2015-06-23
[WIP] Enh: Remove bottle and add it to dependency

Seb-Solon:
Bottlewebui is  used in the webui module on master branch. Bs3 branch remove import and ship it's own bottle in libs.

It is also used is mod-ws-arbiter. I think I will just import bottle instead and add it to requirements (not sur because it depends on Shinken that depends on bottle)

If you agree with that I will commit that to get the compatibility.

If you think about another module that use bottle webui tell me.

Edit : https://github.com/search?l=python&q=+shinken.webui.bottlewebui&type=Code&utf8=%E2%9C%93


gst:
attention: the bottle version included in the source is an old old one not anymore compatible (for quite a long time) with the latest bottle versions released..  ;)

PR #115, opened 2015-07-03
Enh: Add a timeout when joining http_thread so that process is not ha…

Seb-Solon:
Hi,

Test_scheduler_init was regularly failing in Travis / Jenkins. The problem comes from http_thread.join() that hang. I simply added a timeout and a try / except to end the thread. 

"bug" is not easy to reproduce. Just launch the test until you can see it hangs.

### DIFF
--- a/alignak/daemon.py
+++ b/alignak/daemon.py
@@ -340,7 +340,14 @@ class Daemon(object):
 
         if self.http_thread:
             logger.info("Joining http_thread ..")
-            self.http_thread.join()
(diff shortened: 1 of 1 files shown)


gst:
+1 for the quick fix ;)

Yeah the root cause of the problem is that the http threads don't seem to (at least always) have a timeout when reading/writing to a client.. and apparently there are sometimes where *something* connect to the http daemon, but don't send anything.. thus leaving an http thread waiting on its client socket..  happily a new http_daemon re-implementation..

PR #117, opened 2015-07-07
Enh: Remove ModulePhases class was not used

Seb-Solon:
Travis is failing for no reason, I think there are some issue with dep or so. Requests seems to crash sometimes
ping @gst?

### DIFF
--- a/alignak/basemodule.py
+++ b/alignak/basemodule.py
@@ -92,16 +92,6 @@ properties = {
 }
 
 
-class ModulePhases:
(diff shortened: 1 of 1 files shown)


gst:
Hummm.. I don't see what this (ok useless) class has anything to do with any Travis failure.. But the class being useless.. +1 :)

Seb-Solon:
That's why i was pinging, I had a broken build with something unrelated. But anyway you will find some other examples :)

PR #119, opened 2015-07-07
Enh: Test - Try to shutdown thread

Seb-Solon:
Monkey patching to try to stop error in Travis test due to http_server cherrypy refusing connection "sometimes"...

I guess @gst will find the real cause by rewritting the server from scratch :)


gst:
> I guess @gst will find the real cause by rewritting the server from scratch :) 

Once scratched there would be only left an almost   un-noticable remind in our spirit.. ;)

PR #124, opened 2015-07-14
Develop+deldiscovery

ddurieux:
Remove discovery functions because will not be integrated to Alignak.
The goal is too to have other softwares add config to surveil


Seb-Solon:
Looks good to me but I woul stash commit because only 3b017cc does not break tests :)

Issue #87, opened 2015-06-16
Pylint : Todo List


Seb-Solon:
* Split files between people.
* Post here if necessary to talk with .

Seb-Solon:
Rule C0111 (docstring) - Pycharm helps a lot to do it. It generate the following header : 
```
        """
        
        :param parameter: 
        :return:
        """
```

Also you can have the following tags : 
```
:type param: [python type or alignak type with alignak.moduleA.moduleB.ClassB]
:rtype: return type, like above
```

Example : 
```
    """
    Sort host and service by last_state_change

    :param s1: A host or service to compare
    :type s1: alignak.objects.schedulingitem.SchedulingItem
    :param s2: Another host or service to compare
    :type s2: alignak.objects.schedulingitem.SchedulingItem
    :return:
      * -1 if s1 > s2
      * 0 if s1 == s2
      * 1 if s1 < s2
    """
```

gst:
we can do that all of us directly in the same branch.. as far as 2 of us don't work / commit on the same file that'll go smoothly : everyone just has to rebase its local branch on the repository one before pushing.. (otherwise you'll be forbidden by git to push if someone else has done it before ; normal git behavior).

Seb-Solon:
Actually it's the purpose of the Pylint branch on the repository :D

gst:
Yeah, but that wasn't explicitly stated before :p now it's done ;)

Seb-Solon:
Update : 

```
      1 E0213
      1 E0611
      1 E1003
      1 F0010
      1 W0632
      2 E1001
      2 I0013
      2 W0104
      3 C0301
      3 W0621
      4 E0602
      4 W0603
      7 W0221
      9 W0631
     10 F0401
     10 R0911
     11 R0923
     13 C0302
     15 I0011
     16 C1001
     16 R0904
     17 W0102
     27 R0915
     28 R0903
     32 R0914
     37 E0203
     37 R0902
     37 W0613
     39 C0330
     48 R0913
     52 R0912
     55 W0511
     64 W0703
     66 W0612
     67 R0201
     87 R0801
    183 W0212
    436 C0103
    454 W0201
    869 E1101
```

I would suggest to remove E1101 and W0201 

```
:no-member (E1101): *%s %r has no %r member*
  Used when a variable is accessed for an unexistent member. This message
  belongs to the typecheck checker.

:attribute-defined-outside-init (W0201): *Attribute %r defined outside __init__*
[...]

PR #126, opened 2015-07-20
Enh: Pylint (W0403)

ddurieux:


Seb-Solon:
Seems legit. I don't know the diff beween this and local import. Is it very different? ping @gst

Seb-Solon:
OK i found some doc : http://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module

I guess absolute are better then :)

gst:
sorry I had not seen (until now) your ping ;)

absolutes and relatives import are both possible, effectively. and there are use case where relatives imports are really/clearly better (not to say absolutely required)..
but we don't have such use cases here.

So always choosing/using the  absolute side/possibility is OK for me, and it's better to always use the same features the same way ; that's simply more consistent also :)

So +1  ;)

Seb-Solon:

PR #129, opened 2015-07-20
Enh pylint w1401

Seb-Solon:
```
:anomalous-backslash-in-string (W1401): *Anomalous backslash in string: '%s'. String constant might be missing an r prefix.*
  Used when a backslash is in a literal string but not as an escape. This
  message belongs to the string_constant checker.
```


Seb-Solon:

PR #131, opened 2015-07-20
Enh pylint w0614

Seb-Solon:
Only warning were in bottlewebui, I skip it as we will rewrite this

```
:unused-wildcard-import (W0614): *Unused import %s from wildcard import*
  Used when an imported module or variable is not used from a 'from X import *'
  style import. This message belongs to the variables checker.

```


Seb-Solon:

PR #132, opened 2015-07-20
Enh pylint w0107

Seb-Solon:
```
:unnecessary-pass (W0107): *Unnecessary pass statement*
  Used when a "pass" statement that can be avoided is encountered. This message
  belongs to the basic checker.

```


Seb-Solon:

PR #133, opened 2015-07-20
WIP : First step to fix #104

### LINKED ISSUE #104: Implement true Alignak reload
I would like for Alignak-arbiter to reload (config) without changing PID.

jremond:
add SIGHUP generation in init script and signal catching in daemon.py

### DIFF
--- a/bin/init.d/alignak
+++ b/bin/init.d/alignak
@@ -504,14 +504,17 @@ do_reload_() {
         if [ $checkrc -ne 0 ]; then
            return 1
         fi
+        pid=$(getmodpid "$mod")
+        # send SIGHUP signal to reload configuration
+        kill -1 $pid
+        rc=$?
+    else
+        # if not the arbiter module, reload == restart
+        do_restart_ $mod
     fi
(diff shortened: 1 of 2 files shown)


Seb-Solon:
For now it's fine. Now we "only" miss the function reload_config (or something like that)  and call it when logging "Reloading configuration"

PS : Pep8 is sad , see https://travis-ci.org/Alignak-monitoring/alignak/jobs/71862174#L905, not a big deal.

bin/init.d/alignak
```diff
@@ -504,14 +504,17 @@ do_reload_() {
         if [ $checkrc -ne 0 ]; then
            return 1
         fi
+        pid=$(getmodpid "$mod")
+        # send SIGHUP signal to reload configuration
+        kill -1 $pid
```
gst: please use signal name (-HUP) to kill

gst:
The question is now : what to do exactly when the HUP is received..

Seb-Solon:
Done with #265

PR #134, opened 2015-07-20
Enh pylint rules

Seb-Solon:
```
:bad-mcs-classmethod-argument (C0204): *Metaclass class method %s should have %s as first argument*
  Used when a metaclass class method has a first argument named differently than
  the value specified in valid-metaclass-classmethod-first-arg option (default
  to "mcs"), recommended to easily differentiate them from regular instance
  methods. This message belongs to the classes checker.

:duplicate-key (W0109): *Duplicate key %r in dictionary*
  Used when a dictionary expression binds the same key multiple times. This
  message belongs to the basic checker.

:abstract-method (W0223): *Method %r is abstract in class %r but is not overridden*
  Used when an abstract method (i.e. raise NotImplementedError) is not
  overridden in concrete class. This message belongs to the classes checker.

:bad-indentation (W0311): *Bad indentation. Found %s %s, expected %s*
  Used when an unexpected number of indentation's tabulations or spaces has been
  found. This message belongs to the format checker.

:reimported (W0404): *Reimport %r (imported line %s)*
  Used when a module is reimported multiple times. This message belongs to the
  imports checker.

:redefine-in-handler (W0623): *Redefining name %r from %s in exception handler*
  Used when an exception handler assigns the exception to an existing name This
  message belongs to the variables checker.

:unpacking-non-sequence (W0633): *Attempting to unpack a non-sequence%s*
  Used when something which is not a sequence is used in an unpack assignment
  This message belongs to the variables checker.

:cell-var-from-loop (W0640): *Cell variable %s defined in loop*
  A variable used in a closure is defined in a loop. This will result in all
  closures using the same value for the closed-over variable. This message
  belongs to the variables checker.

:pointless-string-statement (W0105): *String statement has no effect*
  Used when a string is used as a statement (which of course has no effect).
  This is a particular case of W0104 with its own message so you can easily
  disable it if you're using those strings as documentation, instead of
  comments. This message belongs to the basic checker.
```

For W0640 I could have ignored it but I found a good workaround here : http://stackoverflow.com/questions/25314547/cell-var-from-loop-warning-from-pylint

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
   - nosetests -xv --process-restartworker --processes=1 --process-timeout=300  --with-coverage --cover-package=alignak
   - coverage combine
   - cd .. && pep8 --max-line-length=100 --exclude='*.pyc' alignak/*
-  - pylint --rcfile=.pylintrc --disable=all --enable=C0111 --enable=W0403 --enable=W0106 --enable=W1401 --enable=W0614 --enable=W0107 --enable=W0141 -r no alignak/*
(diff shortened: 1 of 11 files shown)


Seb-Solon:

PR #136, opened 2015-07-21
Fix E1103 pylint error in trigger_functions

titilambert:


Seb-Solon:

PR #137, opened 2015-07-21
Fix W0141 pylint error

titilambert:


Seb-Solon:

PR #140, opened 2015-07-21
Fix W0231 pylint error

titilambert:
_init__ method from base class '******' is not called


Seb-Solon:
MIssing docstring bro :D

PR #141, opened 2015-07-21
Pylint w0231 w1201 c0325

titilambert:


Seb-Solon:

PR #143, opened 2015-07-23
Typo fixes + load namespace modules

ddurieux:


alignak/modules/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+# Alignak is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Alignak.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+# This file incorporates work covered by the following copyright and
```
Seb-Solon: Should be removed until line 27

Seb-Solon:
This was a duplicate (partial) of #97 that's why we can't mergi this now.

gst:
So, what should be done here ? close or merge (with rebase before I guess) ??

ddurieux:
Update commits to have only part to load the modules.

So we keep $PLUGINSDIR$ for external modules + try to load too in the .../site-packages/alignak/modules/ folder

PR #145, opened 2015-07-24
Pylint

titilambert:


alignak/objects/trigger.py
```diff
@@ -96,6 +96,7 @@ def compile(self):
         """
         self.code_bin = compile(self.code_src, "<irc>", "exec")
 
+    @staticmethod
```
Seb-Solon: Trigget are still exec with trigger instance as far as I remember.

Seb-Solon:
My second comment explinas why test fail IMH

titilambert:
This one seems better :)

Seb-Solon:

PR #147, opened 2015-07-24
Fix clean stop bottle of daemons when press CTRL+C

ddurieux:


alignak/http_daemon.py
```diff
@@ -243,6 +244,10 @@ def get_socks_activity(self, socks, timeout):
     # We are asking us to stop, so we close our sockets
     def stop(self):
         self.stop_requested = True
+        for t in self.threads:
+            while t.is_alive():
+                t.join(5)
```
Seb-Solon: you may be stuck forever in this loop.

alignak/http_daemon.py
```diff
@@ -255,18 +260,17 @@ def run(self):
...
         # Keep a list of our running threads
-        threads = []
         logger.info('Using a %d http pool size', nb_threads)
         while not self.stop_requested:
             # We must not run too much threads, so we will loop until
             # we got at least one free slot available
             free_slots = 0
             while free_slots <= 0 and not self.stop_requested:
-                to_del = [t for t in threads if not t.is_alive()]
+                to_del = [t for t in self.threads if not t.is_alive()]
```
Seb-Solon: Looks like this was always empty before.

alignak/http_daemon.py
gst: I agree,

more specifically it will be stuck if any client request is actually, for whatever reason, blocked / suspended.

I'd more safely do : 
```
end = time.time() + 15    # say give 15 seconds to wait all of them
while time.time() < end: 
    for t in tuple(self.threads):
        if not t.is_alive():
            t.join()
            self.threads.remove(t)
        if not self.threads:
            break
        time.sleep(1)
```

gst: and eventually if we would had a way to interrupt / stop such an http client thread that would be the best, but it's something else.. another PR ;)

gst: > I'd more safely do :

But I'd certainly add a warning if some of those client threads are still alive after this while loop. 'cause it's not normal, in the usual case, for a client thread to stay connected that long.

gst:
Please update to this form : 

```
end = time.time() + 15    # say give 15 seconds to wait all of them
logger.info("Joining my http client threads ..")
while time.time() < end: 
    for t in tuple(self.threads):
        if not t.is_alive():
            t.join()
            self.threads.remove(t)
        if not self.threads:
            break
        time.sleep(1)
if self.threads:
    logger.warning("Some (%s) http threads remain alive: %s", len(self.threads), self.threads)
```

otherwise you are doing a wait up to 5 seconds on each thread, if there had 10 of them and each would be "blocked" then this would delay the stop of 50 seconds..

ddurieux:
Bottle is removed, no need this

PR #151, opened 2015-07-31
Enh: Remove id in variable name and attribute

Seb-Solon:
Id is a builtin variable. Next step for pylint is type and fews thing. 

This one is kinda tricky so I did a pull request separated


Seb-Solon:

PR #155, opened 2015-08-04
Enh Pylint : w0622

Seb-Solon:
```
:redefined-builtin (W0622): *Redefining built-in %r*
  Used when a variable or function override a built-in. This message belongs to
  the variables checker.
```

### DIFF
--- a/alignak/basemodule.py
+++ b/alignak/basemodule.py
@@ -57,7 +57,7 @@ import os
 import signal
 import time
 import traceback
-from re import compile
+import re
 from multiprocessing import Queue, Process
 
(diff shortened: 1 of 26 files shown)


alignak/basemodule.py
```diff
@@ -57,7 +57,7 @@
 import signal
 import time
 import traceback
-from re import compile
+import re
```
ddurieux: But you import all re instead only re.compile...

Seb-Solon: I don't think there is a significant difference in memory IMO.

Seb-Solon:

PR #156, opened 2015-08-05
Assignment replaced with augmented assignment

ddurieux:


Seb-Solon:
Is this a pylint rule? If not, how did you get all ocurrences of this pattern?

ddurieux:
It's when I use 'inspect code' function of PyCharm ;)
Not found rules for pylint rule for that

PR #160, opened 2015-08-05
[wip] Force dict to display as string in logger

ddurieux:


Seb-Solon:
Str(dict) is different from dict.__str__() ?

ddurieux:
Yes because dictionary has not str() function

gst:
??

```
>>> {}.__str__()
'{}'
```

if its not displayed as a string in a logger call, as/like what would it be displayed ??

ddurieux:
Ok I close

Issue #164, opened 2015-08-10
daterange test failure currently


gst:
```
gstarck@taf 10:07 $ nosetests -xv test_dateranges.py:TestDataranges.test_weekdaydaterange_start_end_time
test_weekdaydaterange_start_end_time (test_dateranges.TestDataranges) ... FAIL

======================================================================
FAIL: test_weekdaydaterange_start_end_time (test_dateranges.TestDataranges)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gstarck/work/work/SFL/Supervision/alignak/test/test_dateranges.py", line 139, in test_weekdaydaterange_start_end_time
    self.assertEqual(start, ret[0])
AssertionError: 1442203200.0 != 1439179200.0
-------------------- >> begin captured stdout << ---------------------
```

to be investigated ..  probably a "timing" issue I'd say at first.

gst:
> probably a "timing" issue I'd say at first.

Effectively, I've changed my date to 1 month ago and then the test succeed :

```
gstarck@taf 09:10 $ date
vendredi 10 juillet 2015, 09:10:14 (UTC-0400)
(alignak) ✔ ~/work/work/SFL/Supervision/alignak/test [develop|…5⚑ 4]
gstarck@taf 09:10 $ nosetests -xv test_dateranges.py:TestDataranges.test_weekdaydaterange_start_end_time
test_weekdaydaterange_start_end_time (test_dateranges.TestDataranges) ... ok

----------------------------------------------------------------------
Ran 1 test in 0.034s

OK
(alignak) ✔ ~/work/work/SFL/Supervision/alignak/test [develop|…5⚑ 4]
```

gst:
working on this..

gst:
@ddurieux 

I give it to you as you initialy wrote that test.. don't know what's going on.. but clearly there is/are some things to correct/fix .. I don't know if it's in the test only, or in daterange code itself, or both..

gst:
and if you set your system date to, for instance, 1st January 2015 then there are even more assertion errors in this test module (test_dateranges.py)..

gst:
@ddurieux 

for you to easier debug, you can launch a loop to check each day of the year (2015 or not) like this : 

```
# assuming you are in alignak/test :

$ cp /bin/date .  && sudo chmod +s date  # make a local date sticky..  
# then :
[...]

PR #165, opened 2015-08-10
Deprecate 2 bad methods

gst:
2 (obvious) methods in less.

NB: The test will fail because of https://github.com/Alignak-monitoring/alignak/issues/164

### DIFF
--- a/alignak/basemodule.py
+++ b/alignak/basemodule.py
@@ -59,6 +59,7 @@ import time
 import traceback
 import re
 from multiprocessing import Queue, Process
+import warnings
 
 import alignak.http_daemon
 from alignak.log import logger
@@ -274,6 +275,11 @@ class BaseModule(object):
         :return: True if has a property, otherwise False
         :rtype: bool
         """
+        warnings.warn(
+            "{s.__class__.__name__} is deprecated, please use "
+            "`hasattr(your_object, attr)` instead. This has() method will "
+            "be removed in a later version.".format(s=self),
+            DeprecationWarning, stacklevel=2)
         return hasattr(self, prop)
 
     def want_brok(self, b):
--- a/alignak/objects/contactgroup.py
+++ b/alignak/objects/contactgroup.py
@@ -256,7 +256,8 @@ class Contactgroups(Itemgroups):
             tmp_cg.already_explode = False
 
         for contactgroup in self.items.values():
-            if contactgroup.has('contactgroup_members') and not contactgroup.already_explode:
+            if hasattr(contactgroup, 'contactgroup_members') and not \
+                    contactgroup.already_explode:
                 # get_contacts_by_explosion is a recursive
                 # function, so we must tag hg so we do not loop
                 for tmp_cg in self.items.values():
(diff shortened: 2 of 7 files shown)


alignak/basemodule.py
```diff
@@ -273,6 +274,11 @@ def has(self, prop):
         :return: True if has a property, otherwise False
         :rtype: bool
         """
+        warnings.warn(
```
ddurieux: Why not use logger.warning?

ddurieux:
There are too many warn in the tests, perhaps fix them before merge ;)

alignak/basemodule.py
gst: À priori, this is the python way for deprecating ;)

gst:
? Related to this change ?

alignak/basemodule.py
ddurieux: Ok ;)

ddurieux:
Yes, many:

/home/travis/build/Alignak-monitoring/alignak/alignak/objects/itemgroup.py:99: DeprecationWarning: Servicegroup is deprecated, please use `hasattr(your_object, attr)` instead. This has() method will be removed in a later version.

and on other files

Seb-Solon:
Imo, you don't need 5 commits for that simple replacement of has in hasattr.

Anyway you should rebase, as test a now working.

gst:
@ddurieux  : sweet, that's what a/the deprecation warning is supposed to do :p 

I'll give a look to remove/update the others usages of those 2 functions.

@Seb-Solon :  that's how deprecation is supposed to happen : first you emit/add a warning in the code you want to deprecate, then - and it could be done in more than one supplementary step - you remove/update the usages of the deprecated thing (possibly you may not found all the usages directly, so the need/usefull of the warning)  ;)
but I could have done all of it in 1 commit (per deprecated function/thing). it's just better like it is :p

and yeah I know I'll rebase :p but somehow later..

Seb-Solon:
>: that's how deprecation is supposed to happen : first you emit/add a warning in the code you want to deprecate, then - and it could be done in more than one supplementary step - you remove/update the usages of the deprecated thing (possibly you may not found all the usages directly, so the need/usefull of the warning) ;)

I was not talking about deprecation. +27 −6 could be a single commit, that's all.

gst:
here it is..

gst:
pep8 fixed

gst:
rebased, should be super green now..

ddurieux:
I'm ok this this, perhaps squash it in only one commit

gst:
not necessary to squash, that's not forth and back commit with dirty changes done and undone ;)

ddurieux:
Ok, so you have do all changes needed?  if yes I merge

gst:
yes,

if I'd have forgotten (not impossible though I don't think there are) to
replace some `has(..)` calls then they will be catched by the deprecation
warning (and still be correctly run).

so .. go :p

2015-08-31 8:58 GMT-04:00 David Durieux <notifications@github.com>:

PR #172, opened 2015-08-13
Fix / enhance properties + rename class name with camalcase convention

ddurieux:
Fix

Rename class name with camalcase convention

Fix pep8

### DIFF
--- a/alignak/check.py
+++ b/alignak/check.py
@@ -52,7 +52,7 @@
 
 """
 from alignak.action import Action
-from alignak.property import BoolProp, IntegerProp, FloatProp
+from alignak.property import BoolProp, IntegerProp, ListProp
 from alignak.property import StringProp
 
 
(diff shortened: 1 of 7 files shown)


Seb-Solon:
What pep8 rule you are applying? As pep8 is already done we may add some other ruless

ddurieux:
Not use pep8 rule, juste re-defined properties in the top tree of classes and rename it because not conventional (__Action) and have error (not work) with __Action.properties.copy()

Seb-Solon:
You say fix pep8 in description. I was wondering what you are fixing. Why __Action is not conventional? __is for private stuff. I could make sense. The fact __Action.properties.copy does nothing is strange though

ddurieux:
pep8 in description is because I had 2 lines between 2 funtions instead 1 line :)
And yes __Action.properties.copy not work, it make error, you can see tests errors here https://travis-ci.org/ddurieux/alignak/jobs/75433260

Seb-Solon:
Ok. It's OK IMO but some test are missing as you fix some properties.

You can enhance https://github.com/Alignak-monitoring/alignak/blob/develop/test/test_end_parsing_types.py. As may not have self.conf.notifications/actions/checks (they are created by sched at run time) you can just create empty one and check type.

Is that OK for you?

ddurieux:
OK, I put this PR in [wip] in waiting add tests ;)

ddurieux:
I don't know how to add a test here :/

gst:
what do you want to test actually ??

otherwise +1

ddurieux:
Ready to review, added tests and fixed default values of properties

alignak/check.py
```diff
@@ -123,9 +110,9 @@ def __init__(self, status, command, ref, t_to_go, dep_check=None, _id=None,
         else:
             self.depend_on_me = [dep_check]
         self.check_time = 0
-        self.execution_time = 0
-        self.u_time = 0  # user executon time
-        self.s_time = 0  # system execution time
+        self.execution_time = 0.0
```
Seb-Solon: Nice one

PR #173, opened 2015-08-13
Enh c0103

Seb-Solon:
Hi,

Same rule but for constant name. It make me realize that we should clean some "global" object variable statement at the end of several files (modulesctx, datamanager, logger...)

### DIFF
--- a/alignak/bin/alignak_reactionner.py
+++ b/alignak/bin/alignak_reactionner.py
@@ -63,12 +63,12 @@ import optparse
 import imp
 try:
     imp.find_module('android')
-    is_android = True
+    IS_ANDROID = True
 except ImportError:
-    is_android = False
+    IS_ANDROID = False
 
 
-if is_android is not None:
+if IS_ANDROID is not None:
     # Add our main script dir
     if os.path.exists('/sdcard/sl4a/scripts/'):
         sys.path.append('/sdcard/sl4a/scripts/')
--- a/alignak/log.py
+++ b/alignak/log.py
@@ -70,17 +70,17 @@ from alignak.brok import Brok
 
 # obj = None
 # name = None
-human_timestamp_log = False
+HUMAN_TIMESTAMP_LOG = False
 
-_brokhandler_ = None
+__brokhandler__ = None
 
 
-defaultFormatter = Formatter('[%(created)i] %(levelname)s: %(message)s')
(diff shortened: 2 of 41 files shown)


alignak/bin/alignak_reactionner.py
```diff
@@ -63,12 +63,12 @@
...
     imp.find_module('android')
-    is_android = True
+    IS_ANDROID = True
 except ImportError:
-    is_android = False
+    IS_ANDROID = False
 
 
-if is_android is not None:
+if IS_ANDROID is not None:
```
ddurieux: Not linked with C0103, but IS_ANDROID is never None because we have defined 4 lignes before True or False

alignak/log.py
```diff
@@ -294,14 +294,15 @@ def error(self, *args, **kwargs):
 
 # --- create the main logger ---
 logging.setLoggerClass(Log)
+# pylint: disable=C0103
 logger = logging.getLogger('Alignak')
 if hasattr(sys.stdout, 'isatty'):
-    csh = ColorStreamHandler(sys.stdout)
+    CSH = ColorStreamHandler(sys.stdout)
```
ddurieux: Why csh is in uppercase?

alignak/property.py
```diff
@@ -198,8 +198,8 @@ def __init__(self, text=None):
         self.unused = True
 
 
-_boolean_states = {'1': True, 'yes': True, 'true': True, 'on': True,
-                   '0': False, 'no': False, 'false': False, 'off': False}
+__boolean_states__ = {'1': True, 'yes': True, 'true': True, 'on': True,
```
ddurieux: Seems better to define this in class BoolProp() because it is used only in this class

alignak/log.py
Seb-Solon: Still a constant because defined in module directly. Actually we should change all this but it was not a simple one

alignak/bin/alignak_reactionner.py
Seb-Solon: Will fix in a next pr

alignak/property.py
Seb-Solon: You right, we fix that in 1hour

alignak/log.py
ddurieux: ok

Seb-Solon:

PR #176, opened 2015-08-14
Enh: Revert datamanager constant rename. Will handle it with log etc

Seb-Solon:
Better to handle all the case with var=obj() in module the same way. Datamanager rename will break a lot of module (webui, LS). Let ignore the error for now.


titilambert:
Oki, we really need a new log engine ...

PR #178, opened 2015-08-18
Enh c0103

Seb-Solon:
Method names

### DIFF
--- a/alignak/external_command.py
+++ b/alignak/external_command.py
@@ -92,357 +92,357 @@ class ExternalCommandManager:
     """
 
     commands = {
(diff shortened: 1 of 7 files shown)


alignak/external_command.py
```diff
@@ -721,7 +721,7 @@ def get_command_and_args(self, command, extcmd=None):
             return None
 
         # Now get the command
-        c_name = elts2[1]
+        c_name = elts2[1].lower()
```
ddurieux: force to be upper or keep it ?

Seb-Solon:

PR #187, opened 2015-08-30
Replace cpickle in brok by ujson because better performances (between…

ddurieux:
… 3 times to 10 times)

Test

Add function to allow ujson or cpickle depends on type

Fix broks type to use ujson

Fix unit test

Fix unit test

Clean

### DIFF
--- a/alignak/brok.py
+++ b/alignak/brok.py
@@ -50,6 +50,11 @@ Brok are filled depending on their type (check_result, initial_state ...)
 
 """
 import cPickle
+try:
+    import ujson
+    UJSON_INSTALLED = True
+except ImportError:
+    UJSON_INSTALLED = False
 
 
 class Brok:
@@ -64,7 +69,10 @@ class Brok:
         self.type = _type
         self._id = self.__class__._id
         self.__class__._id += 1
-        self.data = cPickle.dumps(data, cPickle.HIGHEST_PROTOCOL)
+        if self.use_ujson():
+            self.data = ujson.dumps(data)
+        else:
+            self.data = cPickle.dumps(data, cPickle.HIGHEST_PROTOCOL)
         self.prepared = False
 
     def __str__(self):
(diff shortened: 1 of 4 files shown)


ddurieux:
More info, on some broks where have only pure data (so no objects), we use ujson to serialize. for performances, it's faster between 3 to 10 times, depends on length (better perfs when have bigger dictionary)

alignak/brok.py
```diff
@@ -78,7 +82,25 @@ def prepare(self):
...
+        :return: True if type in list allowed, otherwise False
+        :rtype: bool
+        """
+        types_allowed = ['unknown_host_check_result', 'unknown_service_check_result', 'log',
+                         'notification_raise', 'clean_all_my_instance_id', 'initial_broks_done',
+                         'host_next_schedule', 'service_next_schedule', 'host_snapshot',
+                         'service_snapshot', 'host_check_result', 'service_check_result']
+        if self.type in types_allowed:
+            return True
+        return False
```
Seb-Solon: Simple with : retun self.type in [...]

requirements.txt
```diff
@@ -5,3 +5,4 @@ pbr
 termcolor==1.1.0
 setproctitle
 ordereddict==1.1
+ujson
```
Seb-Solon: What about packages of ujson? I'm not sure but I think it's not standard lib.

ddurieux: Packaged in deb / rpm you mean ?

Seb-Solon: Yeah. You know the more dep we have the harder to install.

ddurieux: Seems not in centos official repo list and it's in debian testing, so not packaged :/

Perhaps do this dependency optional and in code use ujson if module installed

gst: This is why I think we should actually "only" focus on "virtualenv"(or --user) installations, system/package(rpm/deb/..) installations should come after.. (though we can keep it in mind).

because **any** system having a proper python2.6 or 2.7 installation (system wide or not) with pip and virtualenv correctly installed will (normally) be able to run alignak.. (if he has the eventual proper dev/headers files/libs required to build some eventual dependency)..

Seb-Solon: Pip may not be available unfortunately that's why we want proper packaging. It's kinda logic to not have pip in production because most of time you have to compile something when installing from pip and gcc is not really recommended is production environments. You also need python.h (python-devel) which is not usually installed.

Another huge con, pip is not made to install configuration files and init script , it only install alignak as a lib, config file shipped are usually not in system path but those are "demo".

gst: > Pip may not be available unfortunately

```
curl https://bootstrap.pypa.io/get-pip.py  > get-pip.py
python get-pip.py
```
what else ? ;)

> Another huge con, pip is not made to install configuration files and init script , it only install alignak as a lib, config file shipped are usually not in system path but those are "demo". 

well, this is relative problems.. as with having pip available : there are certainly solutions.
What I wanted to make more clear: I'm not against if you take consuming time to create deb/rpm/pkgs/.. but that will be less time usable for you to do other things.

gst: >  It's kinda logic to not have pip in production because most of time you have to compile something when installing from pip and gcc is not really recommended is production environments.

it's only a system design decision. cause then you say rolling release distributions are not production usable they are used also in production. 

I was telling there are solutions above, here it is :
Now, with the latest pip version (get-pip) there can have binary wheels pip packages built so that you don't need to compile anything on the production server ..
am I wrong ?

PR #188, opened 2015-09-03
Fix: Test for daterange

Seb-Solon:
For real this time.


ddurieux:
Tests not pass on my computer.
If time.time() get the datetime defined in freezegun, utc or a tz there is no difference...
So I think there is another problem, will see why

Seb-Solon:
The tz parameter given in freezegun seems to be useless. It only parses the given date and consider it as utc

PR #189, opened 2015-09-06
Rewrite httpd

Seb-Solon:
Hi,

Please start the review as it is a __big__ change. I need to do the following : 

* Add test case in full_tst.py 
* Add link to cherrypy doc in some docstring to help understanding.
* Rebase commit and squash some (knitting time)
* Add a parameter to link cherrypy logging with alignak one?
* Add deprecation warning for http_client / http_daemon ?

This is a big first step for HTTP server cleaning. A second step is to go though all exposed method and see if they are really needed. I think some of them could me removed (duplication) or move in a specific daemon interface. 

You will find the following in the pull request : 
* HTTP interfaces previously defined in daemon, satellite, arbiterdaemon ... are now in http/*_interface.py
* HTTP interfaces methods use cherrypy decorators to be registered (see expose). We have now "real" cherrypy app, much easier to read IMHO
* Drop wsgi ref support : we will only use cherrypy now. Specific code regarding backend genericity and backend configuration has been removed. Wsgiref is a simple server, cherrypy has more features so it's a pain to try to used them identically.
* HTTP serialization is a bit different when posting : zlib is made last to send zlib content-type. Pickling is made on json value, keys remain unchanged. This lead to a clean/more standard HTTP communication (using content-type headers)
* Http daemon and http client are now in alignak.http module
* Remove bottle : cherrypy is also used as a framework 
* Http daemon cleaning (linked to wsgi and bottle removal)
* Devel reference update


alignak/daemons/arbiterdaemon.py
```diff
@@ -764,19 +604,20 @@ def setup_new_conf(self):
...
-        self.conf = conf
-        for arb in self.conf.arbiters:
-            if (arb.address, arb.port) == (self.host, self.port):
-                self.me = arb
-                arb.is_me = lambda x: True  # we now definitively know who we are, just keep it.
-            else:
-                arb.is_me = lambda x: False  # and we know who we are not, just keep it.
+        with self.conf_lock:
+            conf = self.new_conf
+            if not conf:
```
ddurieux: Is not be better to do directly :
```if not self.new_conf:```

Seb-Solon: I didn't had a look at the algorithm. I only added a lock to prevent other thread for resetting conf while setting up. Do yeah,this part could be rewritten maybe in another PR :)

Issue #91, opened 2015-06-16
Remove irrelevant files


Seb-Solon:
Remove some useless file. Go through file and see if we can remove them : 

* contrib
* cli => To remove
* for_fedora => See #174 
* libexec => To move ton alignak-contrib org and copy the one needded in test
* manpage => See #174 (is pbr/pip able ot do that)
* inventory => To remove
* external_commands => contrib
* modules => Remove, done by #90 
* media => To remove
* var => module

PR #190, opened 2015-09-07
Enh: Remove irrelevant files #91

Seb-Solon:
NB libexec has been moved to https://github.com/Alignak-monitoring-contrib/alignak-trash-plugins


test/requirements.txt
```diff
@@ -10,3 +10,4 @@ pylint==1.4.1
 pep8==1.5.7
 pep257
 freezegun
+-e git+https://github.com/Alignak-monitoring/mod-example.git#egg=mod-example
```
gst: Bad: commit message says it removes irrelevant files but obviously in fact it does more than that !!

also in test_external_mapping + test_module_as_package + test_modulemanager

Please don't do that !!

Seb-Solon:
I had to fix tet while removing file.  basiaclly it was replacing dummy module with mod example and libexec file

PR #198, opened 2015-09-16
Enh c0103

Seb-Solon:
I updated pylintrc twice to add 10 chars to allowed attribute and function names

### DIFF
--- a/alignak/log.py
+++ b/alignak/log.py
@@ -216,19 +216,19 @@ class Log(logging.Logger):
         # Todo : Do we need this now we use logging?
         return handler.stream.fileno()
 
-    def set_human_format(self, on=True):
+    def set_human_format(self, human=True):
         """
         Set the output as human format.
 
-        If the optional parameter `on` is False, the timestamps format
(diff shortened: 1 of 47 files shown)


alignak/log.py
```diff
@@ -216,19 +216,19 @@ def register_local_log(self, path, level=None, purge_buffer=True):
         # Todo : Do we need this now we use logging?
         return handler.stream.fileno()
 
-    def set_human_format(self, on=True):
+    def set_human_format(self, boolean=True):
```
ddurieux: I think it's not a good idea to name variable 'boolean', perhaps name it 'human'

alignak/macroresolver.py
```diff
@@ -138,7 +138,7 @@ def init(self, conf):
         # Try cache :)
         # self.cache = {}
 
-    def _get_macros(self, s):
+    def _get_macros(self, string):
```
ddurieux: Same think, not name the variable 'string', but something like 'chain' or 'stringvalue'

Seb-Solon:

PR #205, opened 2015-09-23
Fix #177: rewrote get_key_value_sequence to generate_key_value_sequences

### LINKED ISSUE #177: Problem with trailing space in the description of a service
Hi,

Here is an example:

```
define host {
    host_name host-foo
    use       template-with-duplicate-foreach-key
    address   www.xxx.yyy.zzz

    # Key from a "duplicate_foreach" instruction.
    _my_key description n1 $(val1)$ $(val2)$, \
            description n2 $(val1)$ $(val2)$, \
            description n3 $(val1)$ $(val2)$,
}
```

In this example, the descriptions of the services will be `"description n1 "`, `"description n2 "` and `"description n3 "` with a trailing space in the name which will trigger some problems for instance in the WebUI. Indeed, the WebUI will have some dead links (error 500) such as `http://fqdn/service/<host>/<description-of-a-service>` (link to a service).

Currently, there is a workaround: just remove the space between the description and the dollar like this:

```
define host {
    host_name host-foo
    use       template-with-duplicate-foreach-key
    address   www.xxx.yyy.zzz

    # Key from a "duplicate_foreach" instruction.
    _my_key description n1$(val1)$ $(val2)$, \
            description n2$(val1)$ $(val2)$, \
            description n3$(val1)$ $(val2)$,
}
```
In this case, no problem, the descriptions will be `"description n1"`, `"description n2"` and `"description n3"`  without the trailing space.

But it should be smarter if the description of a service was systematically striped, ie with `desc = desc.strip()` somewhere in the code.

Regards.

gst:
duplicate_foreach feature fixed & implementation simplified.

### DIFF
--- a/alignak/objects/service.py
+++ b/alignak/objects/service.py
@@ -76,11 +76,16 @@ from alignak.objects.item import Items
 from alignak.objects.schedulingitem import SchedulingItem
 
 from alignak.autoslots import AutoSlots
-from alignak.util import strip_and_uniq, format_t_into_dhms_format, to_svc_hst_distinct_lists, \
-    get_key_value_sequence, GET_KEY_VALUE_SEQUENCE_ERROR_SYNTAX,\
-    GET_KEY_VALUE_SEQUENCE_ERROR_NODEFAULT, GET_KEY_VALUE_SEQUENCE_ERROR_NODE,\
-    to_list_string_of_names, to_list_of_names, to_name_if_possible, \
-    is_complex_expr
+from alignak.util import (
+    strip_and_uniq,
+    format_t_into_dhms_format,
+    to_svc_hst_distinct_lists,
+    generate_key_value_sequences,
+    to_list_string_of_names,
+    to_list_of_names,
+    to_name_if_possible,
+    is_complex_expr,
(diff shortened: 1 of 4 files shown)


gst:
+240 −296 
 Conversation 0   Commits 1   Files changed 4

Not so bad, not to say really good, given there are added test cases :)

alignak/objects/service.py
```diff
@@ -923,65 +928,60 @@ def duplicate(self, host):
...
+
+        for key_value in key_values:
+            key = key_value['KEY']
+            # Maybe this key is in the NOT list, if so, skip it
+            if key in not_keys:
+                continue
+            new_s = self.copy()
+            new_s.host_name = host.get_name()
+            if self.is_tpl():  # if template, the new one is not
+                new_s.register = 1
```
Seb-Solon: as it's a bool prop i would put a bool here. Strange that there was no bug before...

gst: was orignally like that, I leave it like that, for later if need to be.

Seb-Solon:
LGTM

PR #206, opened 2015-09-23
Fix #184: remove (not anymore supported) android "support"

### LINKED ISSUE #184: remove android support
I'd like to clean the project of any reference to android ; which is anyway not fully supported - if something would work at all..

any one wants to raise some cons about this ?

gst:


gst:
what's left is some config files with android in their name or content..

removing all that also as it also seems concerned.

Issue #219, opened 2015-09-25
code coverage broken (felt from ~65% to ~18%)


gst:
since some time already..

gst:
first coverage which felt : https://coveralls.io/builds/3506360

gst:
related to #222

Seb-Solon:
Fixed with namespace drop

PR #221, opened 2015-09-25
Fix coverage prob

gst:
This should fix the problem with coverage only at 18%.

Don't understood exactly the reason of this but it's apparently somehow linked to modules manager..
EDIT: and also, apparently, the namespace feature ...


Seb-Solon:
What happen to this specific test? Are we going to run it? Mod-example was use for the modules manager, it was replacing the dummy_modules that we removed.

gst:
------------------------------------------------------------------------
TOTAL                                      15610   5225    67%   
----------------------------------------------------------------------

gst:
One thing at a time ;)

gst:
> What happen to this specific test?

it's currently disabled.

>  Are we going to run it?

sure, if we find how to run it without having such problems we had ;)
but that's another PR / correction imho..

gst:
will fix #219

Seb-Solon:
You are here favoring coverage accuracy over a functional test. I'n thinking if it's a good idea or not actually. I don't really have opinion for now. @ddurieux any thought?

gst:
having coverage go from 18% back to our initial 65% at the "cost" of disabling (temporarily, until we find a fix) a **single** test isn't a favor, it's the right thing to do, imho ;)

gst:
@ddurieux, any advice  ?

gst:
ping

Seb-Solon:
I'm not sure this works actually. I tried in another branch to use this code and launch the skipped test with python with no luck. When I remove the python line (so no diff) the coverage was still bad on coveralls. So I'm kinda wondering if this was a "lucky" try or something. Really strange.

You can have a look here : https://github.com/Alignak-monitoring/alignak/pull/234

gst:
had not seen the other PR ;)

gst:
well, I'm not sure you've looked at the right place..

here is the coverage of this PR : https://coveralls.io/builds/3672587 ; which effectively jump back to 64% :

```
#939    develop 64.23   Remove test dependancy on Alignak-monitoring/mod-example.git doesn't appear to be necessary/required at all.    99b4df8fff4fb97b59ae7b6238062f4f?size=18&default=identicongst   PULL #221   25 Sep 2015 travis-ci
```

Seb-Solon:
Yes but mine (which is the same actually) deos not jump back to 64% so I was wondering if your was not a "lucky" coveralls build.

gst:
> Yes but mine (which is the same actually)

almost the same, you miss :

--e git+https://github.com/Alignak-monitoring/mod-example.git#egg=mod-example

which I remember was necessary to be removed in order for the coverage to be fixed..

Seb-Solon:
Yeah but it makes no sense taht coverage try to get cover on this file whitout the test right?. As far as I remember only the modules manager test import all module in its path. I don't get which test get the mod example reached by import.

gst:
> Yeah but it makes no sense taht coverage try to get cover on this file whitout the test right?.

it's not linked to the fact that coverage tries, or not, to cover this package.
It's linked to the fact that it's namespace package(s) and that by being installed then this makes apparently, simply, makes coverage to miss the real package. I should have told that since the start, I agree.

(That would certainly deserves another PR where we try to understand and fix all this mess about coverage, namespace packages, and so on.. and we never know it could be a bug with coverage too..)

gst:
would be great if we should be notified when a branch is no longer mergeable in master because another one has been merged..
i'll rebase..

Seb-Solon:
Coverage is fixed as we remove the namespace feature right?

gst:
good question / but it should be given what we've seen..

gst:
```
The command "if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then ./test/test_all_setup.sh; fi" exited with 1.
```

I find/think that we should not test the setup in travis, I find it counter-producive, travis is primarly used to test the features of a software not how good, or not, it is installed..

we should only test the setup when we generate a new release / version IMHO..

ddurieux:
It's more stable in branch _develop+enhanceunittests_ #310

PR #227, opened 2015-10-02
[wip] Add gzip compression of data of all responses (in cherrypy)

ddurieux:
Clean


Seb-Solon:
All response given by http_daemon to http_client on GET method will be compress by gzip

PR #234, opened 2015-10-07
[WIP] Fix coverage prob (2)

Seb-Solon:
I've added the skipped test after the coverage is done. This way coverage is working and we keep all tests running. 

I had to put the test example back as I'm running the test that need at least one loaded module.

Duplicate of #221


gst:
not sure you've looked at the right place..

here is the coverage of this PR : https://coveralls.io/builds/3672587  ; which effectively jump back to 64% :

```
#939	develop	64.23	Remove test dependancy on Alignak-monitoring/mod-example.git doesn't appear to be necessary/required at all.	99b4df8fff4fb97b59ae7b6238062f4f?size=18&default=identicongst	PULL #221	25 Sep 2015	travis-ci
```

Issue #36, opened 2015-05-27
clean the README.rst


gst:
bad references missed during rename, and content should be changed regarding installation and so..

PR #235, opened 2015-10-18
Enh: Clean README and move to docs #36

Seb-Solon:


README.rst
```diff
@@ -13,10 +13,8 @@ Welcome to the Alignak project.
    :target: https://gitter.im/Alignak-monitoring/alignak?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
 
 
-Alignak is a modern - almost 100% Nagios compatible - monitoring framework,
-written in Python.
-Its main goal is to give users a flexible architecture for
-their monitoring system that is designed to scale to large environments.
+Alignak project is a monitoring framwork based on Shinken who tend to follow OpenStack standards and integrate with it.
+Its main goal is to give users a flexible architecture for  their monitoring system that is designed to scale to large environments.
```
gst: I think we should also state that another of its main goal (if not the most important one) is to be **stable, reliable, and correct**.. which is also one (of the main) reason we forked.. isn't it ?

PR #238, opened 2015-10-30
Fix/enh: rollback namespace package feature

gst:
namespace feature has proved to be difficult to get right in all possible situation, like in virtualenv, in normal setup.py install or --user install, or in pip install. The main problems are:

* Install with pip drop file alignak/__init__.py when install and drop folder alignak/modules where has only __init__.py file inside it
* The external modules has some problems with installation, like have the alignak folder structure and so overwrite files (like alignak/__init__.py, alignak/modules/__init__.py).

Others projects have encoutered the same difficulties (https://specs.openstack.org/openstack/oslo-specs/specs/kilo/drop-namespace-packages.html) and also stepped back using that feature for now.

### DIFF
--- a/alignak/__init__.py
+++ b/alignak/__init__.py
@@ -48,5 +48,4 @@ This file has to be as small as possible in order to namespace to work.
 """
 from . import shinken_import_hook
 
-
-__import__('pkg_resources').declare_namespace(__name__)
+from .version import VERSION as __version__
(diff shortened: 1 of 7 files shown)


gst:
this gonna requires to adapt / clean the modules_manager code

gst:
This one is the more "easy" and doesn't change a lot of things.

Ok for me for merge :)

Seb-Solon:
I think this conflicts with the modules manager rework

gst:
> I think this conflicts with the modules manager rework

which ever will be merged first and then we'll see if there are merge conflicts or any new problem ?

PR #240, opened 2015-10-30
Simplify modules manager

gst:
Simplify modules manager. So we:
* Remove modules_dir and load in a folder
* Will be able to load modules declared like before, but load packages (so can be installed by pip) with name like alignak_module_*

So we remove the compatibility with shinken modules, but we will fork most important / used and packages them properly and diffused through pip.

You can see the code of packages working with this feature:
* mod-example: https://github.com/Alignak-monitoring/mod-example/tree/rollback_namespace_package_feature
* mod-alignakbackend: https://github.com/Alignak-monitoring-contrib/mod-alignakbackend
* mod-nsca: https://github.com/Alignak-monitoring-contrib/mod-nsca


gst:
This one (compared to https://github.com/Alignak-monitoring/alignak/pull/238) is definitievely more touchy/risky..  

I've already commented some parts (more for stylistic concerns than anything else..)

gst:
+1 for merge...

alignak/basemodule.py
```diff
@@ -363,6 +363,14 @@ def set_proctitle(self, name):
         """
         setproctitle("alignak-%s module: %s" % (self.loaded_into, name))
 
+    def main(self):
```
Seb-Solon: +1 !

alignak/modulesmanager.py
```diff
@@ -117,145 +90,56 @@ def set_max_queue_size(self, max_queue_size):
...
-                if uniform_module_type(module.properties['type']) == module_type:
-                    self.modules_assoc.append((mod_conf, module))
-                    break
-            else:  # No module is suitable, we emit a Warning
-                logger.warning("The module type %s for %s was not found in modules!",
-                               module_type, mod_conf.get_name())
+        for mod_conf in mod_confs:
+            try:
+                module = importlib.import_module(mod_conf.module_type)
+                self.find_module_properties_and_get_instance(module, mod_conf.module_type)
```
Seb-Solon: Are you mixing module_type and module name on purpose (see the name variable in find_module_properties_and_get_instace) ?

Seb-Solon: It's a bit confusing IMO, I was expecting a function with a get to return something. Here its modifies module variable actually. Maybe an old Java way of thinking :)

alignak/objects/config.py
```diff
@@ -1219,7 +1215,7 @@ def early_arbiter_linking(self):
             self.arbiters = ArbiterLinks([arb])
 
         # Should look at hacking command_file module first
-        self.hack_old_nagios_parameters_for_arbiter()
+        # self.hack_old_nagios_parameters_for_arbiter()
```
Seb-Solon: If you comment this then you can also remove the old properties dict right?

setup.cfg
```diff
@@ -13,8 +13,7 @@ packages = alignak
 data_files =
     var/log/alignak =
     var/run/alignak =
-    var/lib/alignak/modules =
-    var/lib/alignak/cli = cli/*
```
Seb-Solon: Nice catch

test/requirements.txt
```diff
@@ -10,5 +10,5 @@ pylint==1.4.4
 pep8==1.5.7
 pep257
 freezegun
--e git+https://github.com/Alignak-monitoring/mod-example.git#egg=mod-example
+-e git+https://github.com/Alignak-monitoring/mod-example.git@rollback_namespace_package_feature#egg=mod-example
```
Seb-Solon: Lets merge the mod example first so that we don't hav eto do this?

test/test_module_autogeneration.py
```diff
@@ -48,6 +48,7 @@
 from alignak_test import *
 
 
+@unittest.skipIf(True, "if you want namedpipe then define it in your config !")
```
Seb-Solon: If you want to drop this feature say it explicitely. This is for Nagios copatibility actually, we can "debate" this point

Seb-Solon:
Drop of shinken modules may involve drop of shinken hook right? @gst is the hook still necessary?

See my comments for the rest

What I can see is that here we wont have modules tests working. We may need to comment some part of the jenkins script because we can't modules anymore

alignak/modulesmanager.py
gst: I agree it's not very clear / could be enhanced in this matter..

gst: > Are you mixing module_type and module name on purpose (see the name variable in find_module_properties_and_get_instace) ?

No. re-read find_module_properties_and_get_instance() docstring ;)
Hint: the alignak "module_type" (there is also the alignak "module_name" effectively, which makes it even more confusing) vs python module name (distinction/difference or not) **is** confusing.. I really wish we could have renamed (sooner than latter) the "module" concept name to something less confusing..

But anyway, the function should be (clarified/reworked/) finished actually, it isn't suffisient **at all** I'm afraid / sure.. (the try/except/else is actually to be able to detect "new-style" vs "old-style" shinken/alignak module but it doesn't do more than that, while it should..)

gst:
> Drop of shinken modules may involve drop of shinken hook right? @gst is the hook still necessary?

I would have said the opposite :s 

there can be some shinken modules (there are actually) relying on some hooks, but I don't know of some hooks relying on some shinken module..

or I don't get you ?

alignak/modulesmanager.py
ddurieux: Perhaps have:
* __module_alias__: the identifier of the module (like a reference) and used to declare this module in alignak arbiter/broker....
* __module__:  the module to import (example: alignak_module_example, alignak_module_backend.arbiter)

Seb-Solon: Ok I got confuse on the module name here actually. This is referring to the python module name not the module_name in the configuration. So yeah here you are trying to import the python module named "module_type" which is Ok for me. The module type in the module property should match the directory module name (so the python module name)

+1 for @ddurieux  sugg

Seb-Solon:
The point of dropping shinken module  is to not use them right? Alignak module would have no occurence of "shinken" in the code so that the python hook is not necessary. Maybe you had something else in mind when saying "drop shinken modules". 

It's almost OK for me, see comment unanswered.

test/test_module_autogeneration.py
ddurieux: it's linked with config.hack_old_nagios_parameters_for_arbiter. 
I think it's not used

alignak/modulesmanager.py
ddurieux: If all is ok with rename:
* module_name -> module_alias
* module_type -> module

I will change in code

Seb-Solon:
Forgot some part in the test install, strange test did not failed.

ddurieux:
In reallity, with this new modules manager, shinken modules are not compatible without little code. That's why we have began fork shinken modules and add them compatible alignak (see contrib repository)

alignak/modulesmanager.py
gst: sorry guys, I missed this thread (as well as others here and there) for some time it seems like..

and I'm ofcourse +1 for your suggestion @ddurieux ..  I would have known that we could go and break up to that point the *modules* not so internals (name/type vs alias/wegotit) so, then I think I would have probably suggested that as well for long..  but somehow in this way : 

```
define module {
    name             blablabla         #  a (unique) name !  
                                     # so to be used/distinguishable within daemon configs..
    python_path      alignak_module_example
                   # or even a sub(package|module) with dot notation so : 
                   # toppackage.subpackage.submodule

    # .. others parameters linked to this blablabla module instance ..
}
```
and I would have used the same name/python_path within the code :)

ddurieux: It has been modified yet and merged, we have now:
* module_alias xxx (name of module to declare in broker, arbiter... conf file) 
* python_name alignak_module_xxx (it's the name of python module)

Example: https://github.com/Alignak-monitoring/alignak-module-example/blob/master/etc/alignak/arbiter_cfg/modules/mod-example.cfg

gst: yeah, see that.

After thinking to it 2 secs : I thought we wanted to keep compatibility with shinken configuration files format.. :s

ddurieux: shinken modules not compatible now. That's why we have begun to fork shinken modules and fix them to be compatible with alignak and publish them to pip

PR #257, opened 2015-11-19
Enhance is_impact renamed in is_impacted. closes #256

### LINKED ISSUE #256: Rename is_impact property
Rename __is_impact__ property (host or service) to __is_impacted__ because is_impact has not the real meaning

ddurieux:


Seb-Solon:
Are you sure of that? From my understanding, it was is this object is an impact for other objected. So the opposite. But this part was confusing so I may make a mistake

ddurieux:
Ok I close

PR #258, opened 2015-11-19
Enh: in debug mode, display command line executed and result (poller,…

### LINKED ISSUE #252: Poller debug mode
When the poller daemon log level is set to DEBUG it should be useful to log the launched commands and the results of the launched commands. It will be helpful to test and improve the monitoring configuration

ddurieux:
… reactionner). closes #252

### DIFF
--- a/alignak/action.py
+++ b/alignak/action.py
@@ -182,6 +182,7 @@ class ActionBase(object):
         self.stdoutdata = ''
         self.stderrdata = ''
 
+        logger.debug("Launch command: %s", self.command)
         return self.execute__()  # OS specific part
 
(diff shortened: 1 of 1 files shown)


alignak/action.py
```diff
@@ -394,6 +395,7 @@ def execute__(self, force_shell=sys.version_info < (2, 7)):
             # http://www.doughellmann.com/PyMOTW/subprocess/ for
             # detail about this.
             try:
+                logger.debug("Launch command: %s", self.command)
```
Seb-Solon: I would suggest to put this line 185 so that you only put once.

```
return self.execute__() # OS specific part
```

command is a attribute, you have access to it at this line

PR #260, opened 2015-11-20
Fix linkify templates for commands and triggers

ddurieux:


Seb-Solon:
What parameters were not linked properly?  It could be great to have a test case to see the difference in the objects

ddurieux:
These two objects don't have management of templates and I think it could be a good idea.
If you're ok with that, I will add tests

Seb-Solon:
well triggers templates, it's hard for me to see the usage. Command why not, I don't have a use case but i'm pretty sure we can have some.

ddurieux:
Yes it's a non-sense

PR #266, opened 2015-12-01
Rename module_name and module_type

ddurieux:
Rename:

* __module_name__ in __module_alias__
* __module_type__ in __module__


alignak/objects/config.py
```diff
@@ -1749,38 +1749,38 @@ def fill_default_satellites(self):
...
-        :type mod_type: str
+        :param module: module type to search
+        :type module: str
         :return: True if mod_type is found else False
         :rtype: bool
         """
         for broker in self.brokers:
             for module in broker.modules:
-                if hasattr(module, 'module_type') and module.module_type == mod_type:
+                if hasattr(module, 'module') and module.module == module:
```
Seb-Solon: module.module is strange but still legit in python

alignak/objects/config.py
```diff
@@ -1749,38 +1749,38 @@ def fill_default_satellites(self):
...
+        :param module: module type to search
+        :type module: str
         :return: True if mod_type is found else False
         :rtype: bool
         TODO: Factorize it with got_broker_module_type_defined
         """
         for scheduler in self.schedulers:
             for module in scheduler.modules:
-                if hasattr(module, 'module_type') and module.module_type == mod_type:
+                if hasattr(module, 'module') and module.module == module:
```
Seb-Solon: Same here

alignak/objects/config.py
ddurieux: it's the 'name' of module to import, have you another string name for this? I haven't idea of a good name for this

ddurieux: New proposition of name: python_name

PR #268, opened 2015-12-02
Rewrite 95% function of util in average_percentile.

ddurieux:
Rewrite function correctly with percentile usage (95 for max and 5 for min).
Be careful, I introduce numpy dependance. I think we will use it because I want to enhance the get_raw_stats and general stats of alignak in the future

### DIFF
--- a/alignak/http/scheduler_interface.py
+++ b/alignak/http/scheduler_interface.py
@@ -25,7 +25,7 @@ import zlib
 
 from alignak.log import logger
 from alignak.http.generic_interface import GenericInterface
-from alignak.util import nighty_five_percent
+from alignak.util import average_percentile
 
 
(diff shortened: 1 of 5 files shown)


Seb-Solon:
Which version did you used? Im just giving a small check for packaging

ddurieux:
I use version 1.10.0 but this function exist since many releases, we can use for example >= 1.8

Seb-Solon:
you mean 1.8 >= ? This is actually debian stable. For Ubuntu / centos it is below (1.6>=).

ddurieux:
1.6? can you test if function numpy.percentile works on this version ? I think yes but to be sure

Seb-Solon:
working with precise

Issue #262, opened 2015-11-28
Roadmap


Seb-Solon:
From previous Roadmap:
- [x] Pylint Code : End with Pylint rules naparuba/shinken#1393   #87 (@Seb-Solon, @titilambert, @gst)
- [ ] Doc enahcement : Add again some details about config objects (@alignak-monitoring)
- [x] Enhancement : Rework some utils functions naparuba/shinken#1493 (@Seb-Solon)
- [ ] Enhancement : Rework logger, differentiate naglog and shinken log: Issue=https://github.com/Alignak-monitoring/alignak/issues/70
    For refs, see naparuba/shinken#1525 and naparuba/shinken#1201
- [ ] Test stabilization: bad_start and full test
- [ ] Pypy compatibility
- [ ] Python 3 compat
- [x] Remove pickle and use Json #30 (@ddurieux)
- [ ] Enhancement: Ensure unicode in all stringProp (@Seb-Solon)
- [ ] Enhancement: Default value for ListProp ([] or [''] or None ?)
- [x] Rename alignak modules parameters : module_name and module_type

No Release date

PR #270, opened 2015-12-03
pylint E1101

ddurieux:
:no-member (E1101): *%s %r has no %r member*
  Used when a variable is accessed for an unexistent member. This message
  belongs to the typecheck checker.


alignak/daemon.py
```diff
@@ -198,6 +199,8 @@ def __init__(self, name, config_file, is_daemon, do_replace, debug, debug_file):
         self.debug = debug
         self.debug_file = debug_file
         self.interrupted = False
+        self.pidfile = None
+        self.http_interface = GenericInterface(object)
```
Seb-Solon: Should be none instead right?

alignak/brok.py
```diff
@@ -106,10 +108,11 @@ def prepare(self):
         # if so, the data is already ok
         if hasattr(self, 'prepared') and not self.prepared:
             if self.use_ujson():
+                # pylint: disable=E1101
                 self.data = ujson.loads(self.data)
             else:
                 self.data = cPickle.loads(self.data)
-            if hasattr(self, 'instance_id'):
+            if hasattr(self, 'instance_id') and self.instance_id is not None:
```
Seb-Solon: if self.instance_id ?

alignak/db.py
```diff
@@ -175,19 +175,3 @@ def create_update_query(self, table, data, where_data):
 
         query = query + query_follow + where_clause
         return query
-
```
Seb-Solon: This is use in some modules, why remove it?

alignak/objects/config.py
```diff
@@ -1344,13 +1346,6 @@ def linkify(self):
         # satellites
         self.realms.prepare_for_satellites_conf()
 
-    def remove_exclusions(self):
```
Seb-Solon: Not used?

alignak/objects/arbiterlink.py
```diff
@@ -120,44 +120,6 @@ def do_not_run(self):
             self.con = None
             return False
 
-    def get_satellite_list(self, daemon_type):
```
Seb-Solon: Not used?

alignak/objects/item.py
```diff
@@ -634,85 +637,6 @@ def del_comment(self, comment_id):
         if c_to_del is not None:
             self.comments.remove(c_to_del)
```
Seb-Solon: Unused?

alignak/objects/resultmodulation.py
```diff
@@ -111,8 +111,6 @@ def pythonize(self):
 
         :return: None
         """
-        # First apply Item pythonize
-        super(Resultmodulation, self).pythonize()
```
Seb-Solon: I think you should keep this

alignak/objects/schedulingitem.py
```diff
@@ -1970,3 +1972,275 @@ def fill_data_brok_from(self, data, brok_type):
         # SchedulingItem in its check_result brok
         if brok_type == 'check_result':
             data['command_name'] = self.check_command.command.command_name
+
```
Seb-Solon: Nice one, it makes more sense

alignak/trigger_functions.py
```diff
@@ -56,8 +56,10 @@
 
 from alignak.misc.perfdata import PerfDatas
 from alignak.log import logger
+from alignak.objects.host import Hosts
+from alignak.objects.service import Services
 
-OBJS = {'hosts': [], 'services': []}
+OBJS = {'hosts': Hosts({}), 'services': Services({})}
```
Seb-Solon: This may break existing trigger I think. To be check

alignak/objects/schedulingitem.py
```diff
@@ -1970,3 +1972,275 @@ def fill_data_brok_from(self, data, brok_type):
...
+        :param status: status to compare. Usually comes from config files
+        :type status: str
+        :return: True
+        :rtype: bool
+        """
+        return True
+
+    def raise_freshness_log_entry(self, t_stale_by, t_threshold):
+        """Raise freshness alert entry (warning level)
+        Format is : "The results of host '*get_name()*' are stale by *t_stale_by*
```
Seb-Solon: I think you have to fix some docstring to replace host by item and say it's a abstract method or something.

alignak/objects/schedulingitem.py
```diff
@@ -1970,3 +1972,275 @@ def fill_data_brok_from(self, data, brok_type):
...
+        :param t_stale_by: time in seconds the host has been in a stale state
+        :type t_stale_by: int
+        :param t_threshold: threshold (seconds) to trigger this log entry
+        :type t_threshold: int
+        :return: None
+        """
+        pass
+
+    def raise_snapshot_log_entry(self, command):
+        """Raise HOST SNAPSHOT entry (critical level)
```
Seb-Solon: here too

alignak/objects/schedulingitem.py
```diff
@@ -1970,3 +1972,275 @@ def fill_data_brok_from(self, data, brok_type):
...
+
+        :return: list containing the service and the linked host
+        :rtype: list
+        """
+        return []
+
+    def get_data_for_event_handler(self):
+        """Get data for an event handler
+
+        :return: list containing a single host (this one)
```
Seb-Solon: item

Seb-Solon:
First look done. Need another one + fix

alignak/daemon.py
ddurieux: No because in the code, we use properties or functions of GenericInterface, so I initialize it with that

alignak/objects/arbiterlink.py
ddurieux: no ;)

alignak/objects/config.py
ddurieux: No ;)

alignak/objects/item.py
ddurieux: Moved in alignak/objects/schedulingitem.py because only used by object inherite schedulingitem

alignak/objects/resultmodulation.py
ddurieux: Keep what part?

gst: The super call ?

Seb-Solon: Yes

ddurieux: I have re-add it, but there is a problem, Item.pythonize() not exist

Seb-Solon: You are right. Actually we have to remove the whole function. Pythonize is done on properties when instanciating the object. So Item.__init__ is already doing the job.

ddurieux: You're right, seems this function never called in this class ;)

I will remove it

alignak/daemon.py
ddurieux: Are you against this ?

alignak/db.py
ddurieux: Good question, I will check it

alignak/daemon.py
Seb-Solon: Well if pylnt complains about tye then we should respect this type. I was more concern about memory consumption and unuesed object in memory

Seb-Solon:
Related to #87  and naparuba/shinken#1393 . Part of the Roadmap #262

Need to update #87 to get the new count. Gonna be a huge step forward.

PR #271, opened 2015-12-17
Enh/Fix: prevent memleak in any module worker who use logging

gst:
The module worker doesn't forward their broks to their parent process / anywhere.

### DIFF
--- a/alignak/worker.py
+++ b/alignak/worker.py
@@ -59,11 +59,11 @@ import traceback
 import cStringIO
 
 
-from alignak.log import logger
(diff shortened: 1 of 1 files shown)


gst:
the left issues with travis are not because of this commit..
as far as I can see they are not new..

Seb-Solon:
Why removing the object inheritance?

gst:
you mean add it ;) 
new style class always better :) as i modified that class i updated that as well, doesn't hurt here as worker instances are not "specially" used (like serialized) afaik..

PR #275, opened 2016-01-06
Enh: Init script - Clean and add fallbacks for lsb functions

Seb-Solon:
Hi,

This is mainly for packaging purpose only. Basically systemd based system may have not all init function so I add a fallback. next step is provide a service file and not this file. 

I also cleaned some code, and set the default value of default file as /etc/default/alignak if the relative path does not work (sometimes /etc/init.d is symlink so /etc/init.d/../default/alignak gives /etc/rc.d/default/alignak)


Seb-Solon:
Also fixes naparuba/shinken#1795

PR #277, opened 2016-01-18
Fixed issue https://github.com/naparuba/shinken/issues/1779

titilambert:


alignak/objects/host.py
```diff
@@ -1133,7 +1133,8 @@ def set_state_from_exit_status(self, status):
             state_code = 'd'
         if state_code in self.flap_detection_options:
             self.add_flapping_change(self.state != self.last_state)
-        if self.state != self.last_state:
+        if self.state != self.last_state and \
+                not(self.state == "DOWN" and self.last_state == "UNREACHABLE"):
             self.last_state_change = self.last_state_update
```
ddurieux: Perhaps it's a good idea to add unit test to test all possibilities, like change UNREACHEABLE -> DOWN, it not restart time since last_change...
I think to report fix in services because it's the same problem

Seb-Solon:
+1 @ddurieux , need some test case on that. Any ideas?

ddurieux:
Fixed in 6ac056f01e7643e0291ca4a721f5a78d26bff300

PR #285, opened 2016-01-29
Enh: - Pylint W0102

Seb-Solon:
```
:dangerous-default-value (W0102): *Dangerous default value %s as argument*
  Used when a mutable value as list or dictionary is detected in a default value
  for an argument. This message belongs to the basic checker.
```


alignak/check.py
```diff
@@ -117,7 +117,10 @@ def __init__(self, status, command, ref, t_to_go, dep_check=None, _id=None,
         self.poller_tag = poller_tag
         self.reactionner_tag = reactionner_tag
         self.module_type = module_type
-        self.env = env
+        if env is not None:
```
ddurieux: Why not keep env={} in line 89 and only have: 
```
if env != {}
    self.env = env
```

PR #287, opened 2016-01-30
Almost done with Pylint

Seb-Solon:
```
:too-many-return-statements (R0911): *Too many return statements (%s/%s)*
  Used when a function or method has too many return statement, making it hard
  to follow. This message belongs to the design checker.

:too-many-lines (C0302): *Too many lines in module (%s/%s)*
  Used when a module has too much lines, reducing its readability. This message
  belongs to the format checker.

```

#262 #87 and naparuba/shinken#1393


alignak/objects/host.py
```diff
@@ -1036,46 +1036,34 @@ def notification_is_blocked_by_item(self, n_type, t_wished=None):
...
-            if self.state == 'DOWN' and 'd' not in self.notification_options:
-                return True
-            if self.state == 'UP' and 'r' not in self.notification_options:
-                return True
-            if self.state == 'UNREACHABLE' and 'u' not in self.notification_options:
-                return True
-        if (n_type in ('FLAPPINGSTART', 'FLAPPINGSTOP', 'FLAPPINGDISABLED')
-                and 'f' not in self.notification_options):
+        if n_type in ('PROBLEM', 'RECOVERY') and (
+                self.state == 'DOWN' and 'd' not in self.notification_options or
```
ddurieux: I think you must add () like
```
(self.state == 'DOWN' and 'd' not in self.notification_options) or
```

Seb-Solon: And is the prioritary operator and actually.

```
In [1]: True and True or False and False
Out[1]: True

```

PR #293, opened 2016-02-26
Fix load dataranges of timeperiod from config files. closes #292

### LINKED ISSUE #292: Timeperiod problem on load
When load a timeperiod like with 2 _monday_, the second is droppped, example:

```
define timeperiod{
    name            us-holidays
        timeperiod_name         us-holidays
        alias                   U.S. Holidays

        monday -1 may           00:00-00:00     ; Memorial Day (last Monday in May)
        monday 1 september      00:00-00:00     ; Labor Day (first Monday in September)
        }
```

ddurieux:
Closes #292.

Fixes a computation problem raised with Daylight Saving Time. Like in october/november, between 1:00 am and 4:00 am, the timestamp is incremented by 4h x 3600s, but the localtime is converted to correct date/time and so it's 4:00am and not 5:00am


ddurieux:
I have found bugs on some configurations

ddurieux:
It's ok for me now ;)

Seb-Solon:
Ok I think the fix should be here : https://github.com/Alignak-monitoring/alignak/blob/develop/alignak/objects/timeperiod.py#L184-L189

The if isinstance is always try I think. We should do for loop on the value list and add key + value for each. 

You case change the parsing for *every* objects which is more dangerous + change the key : value assignment to 'key value: '' in some cases which could be a side effect.

What do you think?

test/_old/test_timeperiods.py
```diff
@@ -527,8 +527,11 @@ def test_issue_1385(self):
             invalid_tm = time.mktime(dt.timetuple())
             self.assertFalse(tp.is_time_valid(invalid_tm))
 
-
-
+    def test_timeperiod_multiple_monday(self):
```
mohierf: You leave this file in the *test/_old* directory ? Why not moving it to the *test* directory ???

mohierf: @ddurieux : no answer for this question?

mohierf:
Looks good to me ... except for two things:
- you should squash to have one only commit
- you should explain clearly in the PR description what is exactly the DST problem that got fixed with the last modifications. I did not saw any issue for this problem ?

ddurieux:
* Yep, I will squash
* with hour changement like in october/november, between 1:00 am and 4:00 am, the timestamp is incremented by 4h x 3600s, but the localtime convert in right date/time and so it's 4:00am and not 5:00am

PR #159, opened 2015-08-05
[wip] Convert brok class to new style class

ddurieux:

Issue #30, opened 2015-05-25
cpickle vs simplejson


ddurieux:
I have seen many performances pages about thi son the web and we can see all result (cpickle > simplesjon and other cpickle < simplejson).

Somebody has made performances tests with both in alignak (currently use cpickle)?

If no, I can try it

Seb-Solon:
well i you want to compare, try to compare the c version of json (usjon I thnk) I will be fair vs cPickle ;)

ddurieux:
I have tested ujson and simplejson and we have problem because have objects in dict to serialize.

For example, I have in a dict
```
'contacts': <shinken.objects.contact.Contacts object at 0x808eaa6d0
```

So why not put in dict only data and the module use the object to assicate data with object. The good point for this is the data to serialize and size will be more small to pass between module via http/https.

Seb-Solon:
We could start with define dumps and loads function for every item (in Item?) so that we can call json on it

ddurieux:
cpickle removed to json in #294

PR #294, opened 2016-04-03
Unlink objects, Remove pickle, reworks items

Seb-Solon:
Hi,

This pull request do the following : 

* Use uuid in all objects instead of id
* Unlink almost all items : attribute of object are no instance of other objects except for a few cases (commandcall / check_period and acknowledgemnt)
* Replace pickle by manual serialization : return a json representation of objects. 
* Drop of base64 / zlib manual serialisation in HTTP and use CherryPy facilities to have json
* Rework class inheritance: instantiation of object is more generic 
* Clean unused functions and class
* Increase CherryPy queue for larger environnment

This will reduce memory leaks because object will be cleaned in memory as they are not linked anymore

Related to #30 #159 #262  naparuba/shinken#1600 naparuba/shinken#1522 naparuba/shinken#1793 naparuba/shinken#1819 naparuba/shinken#1828 naparuba/shinken#1836


alignak/acknowledge.py
```diff
@@ -101,34 +99,16 @@ def __init__(self, ref, sticky, notify, persistent,
...
         """
-        cls = self.__class__
-        self._id = state['_id']
-        for prop in cls.properties:
-            if prop in state:
-                setattr(self, prop, state[prop])
-        # If load a old ack, set the end_time to 0 which refers to infinite
-        if not hasattr(self, 'end_time'):
-            self.end_time = 0
+        return {'uuid': self.uuid, 'ref': self.ref, 'sticky': self.sticky, 'notify': self.notify,
```
ddurieux: The self.ref is the Schedulingitem (host for example), so you put in the serialization a complete object. Not sure we can ujson this dict. Perhaps need to add in this the object name + the UUID

alignak/action.py
```diff
@@ -101,41 +102,48 @@ def no_block_read(output):
...
-        """Set Action._id to the maximum of itself and _id
+    def __init__(self, params=None):
+        super(ActionBase, self).__init__(params)
+        self.creation_time = time.time()
+        self.fill_default()
 
-        :param _id: action id to compare (from a previous run usually)
-        :type _id: int
-        :return: None
+    def get_id(self):
```
ddurieux: Perhaps simplify and remove this function and update the 2 lines call this function and use object.uuid instead object.get_id()

alignak/brok.py
```diff
@@ -48,34 +48,43 @@
...
-        self.instance_id = None
-        self.__class__._id += 1
-        if self.use_ujson():
-            # pylint: disable=E1101
-            self.data = ujson.dumps(data)
+    def __init__(self, params):
+        self.uuid = params.get('uuid', uuid.uuid4().hex)
+        self.type = params['type']
+        self.instance_id = params.get('instance_id', None)
+        # Again need to behave diffrently when un-serializing
```
ddurieux: Typo: _differently_ instead _diffrently_

alignak/brok.py
```diff
@@ -97,7 +106,7 @@ def id(self, value):  # pylint: disable=C0103
         """
         warnings.warn("Access to deprecated attribute id of %s class" % self.__class__,
                       DeprecationWarning, stacklevel=2)
-        self._id = value
+        self.uuid = value
 
     def prepare(self):
         """Unpickle data from data attribute and add instance_id key if necessary
```
ddurieux: Fix descr: _Un-serialize_ instead _Unpickle_

alignak/brok.py
```diff
@@ -107,26 +116,10 @@ def prepare(self):
...
         if hasattr(self, 'prepared') and not self.prepared:
-            if self.use_ujson():
-                # pylint: disable=E1101
-                self.data = ujson.loads(self.data)
-            else:
-                self.data = cPickle.loads(self.data)
+            try:
+                self.data = unserialize(self.data)
+            except AlignakClassLookupException:
+                raise
```
ddurieux: it could be a good idea to add error text to know what is the problem

alignak/commandcall.py
```diff
@@ -83,39 +76,53 @@ class CommandCall(DummyCommandCall):
...
+                self.poller_tag = params.get('poller_tag', 'None')  # from host/service
+                self.reactionner_tag = params.get('reactionner_tag', 'None')
+                self.module_type = self.command.module_type
+                self.enable_environment_macros = self.command.enable_environment_macros
+                self.timeout = int(self.command.timeout)
+                if self.valid and params.get('poller_tag', 'None') == 'None':
+                    # from command if not set
+                    self.poller_tag = self.command.poller_tag
+                # Same for reactionner tag
+                if self.valid and params.get('reactionner_tag', 'None') == 'None':
```
ddurieux: Why not use directly _self.reactionner_tag_ instead use: 
`params.get('reactionner_tag', 'None')`. It will be more simple ;)

alignak/daemon.py
```diff
@@ -1042,7 +1046,7 @@ def check_for_system_time_change(self):
 
         return difference
 
-    def compensate_system_time_change(self, difference):  # pylint: disable=R0201
+    def compensate_system_time_change(self, difference, timeperiods):  # pylint: disable=R0201,W0613
```
ddurieux: _timeperiods_ argument not used in this function, so why add it?

alignak/objects/config.py
```diff
@@ -2420,7 +2520,7 @@ def cut_into_parts(self):  # pylint: disable=R0912
 
             # we need a deepcopy because each conf
             # will have new hostgroups
-            cur_conf._id = i
+            cur_conf.uuid = i
```
ddurieux: i = int, not a uuid

alignak/objects/contact.py
```diff
@@ -72,15 +73,15 @@ class Contact(Item):
...
+        'host_notification_period': StringProp(default='', fill_brok=['full_status']),
+        'service_notification_period': StringProp(default='', fill_brok=['full_status']),
         'host_notification_options': ListProp(default=[''], fill_brok=['full_status'],
                                               split_on_coma=True),
         'service_notification_options': ListProp(default=[''], fill_brok=['full_status'],
                                                  split_on_coma=True),
         # To be consistent with notificationway object attributes
-        'host_notification_commands': ListProp(fill_brok=['full_status']),
-        'service_notification_commands': ListProp(fill_brok=['full_status']),
+        'host_notification_commands': ListProp(fill_brok=['full_status'], default=[]),
```
ddurieux: To have same arguments order like other, define first _default_ and after _fill_brok_

alignak/brok.py
Seb-Solon: We can't log at this level without reworking the logging in Alignak.

Log depend en brok (Brokhandler) and Brok will dep on will if you want to log here. That why I raise the exception and pray for a try except in a higher level :)

alignak/acknowledge.py
Seb-Solon: Nice catch

alignak/commandcall.py
Seb-Solon: Well I only copy-pasted the previous behavior. It's also legit for poller_tag

alignak/objects/config.py
Seb-Solon: damn you are right, I have tu uuid here and put it in the dict :)

alignak/objects/notificationway.py
```diff
@@ -103,6 +105,33 @@ class NotificationWay(Item):
...
+                # We recreate the object
+                setattr(self, prop, new_list)
+                # And remove prop, to prevent from being overridden
+                del params[prop]
+        super(NotificationWay, self).__init__(params)
+
+    def serialize(self):
+        res = super(NotificationWay, self).serialize()
+
+        for prop in ['service_notification_commands', 'host_notification_commands']:
```
ddurieux: To simplify, perhaps pass this list in argument of .serialize() function because we 'duplicate' these 5 lines in many files

alignak/objects/notificationway.py
```diff
@@ -352,7 +385,7 @@ def new_inner_member(self, name=None, params=None):
         :return: None
         """
         if name is None:
-            name = NotificationWay._id
+            name = 'Generated_notificationway_%s' % uuid.uuid4().hex
```
ddurieux: We initialize this class with init of Item class, so prefer use _self.uuid_ no?

alignak/objects/service.py
```diff
@@ -1471,13 +1346,23 @@ def apply_implicit_inheritance(self, hosts):
...
         """Wrapper to loop over services and call Service.fill_daddy_dependency()
 
         :return: None
         """
         for service in self:
-            service.fill_daddy_dependency()
+            if service.host and service.host_dependency_enabled:
+                host = hosts[service.host]
+                service.act_depend_of.append(
+                    (service.host, ['d', 'u', 's', 'f'], 'network_dep', '', True)
```
ddurieux: I think it's not right options: 
`['d', 'u', 's', 'f']` is values of host, not service

Seb-Solon: I only adapted this from tps://github.com/Alignak-monitoring/alignak/blob/develop/alignak/objects/service.py#L397-L418

We may think later if this is legit or not

alignak/objects/notificationway.py
Seb-Solon: We are in Notificationways, no a Notificationway. Here we create a new object. Still there is a small mistake here in my opinion

ddurieux:
Reminder list of problems:
* Unserialize in scheduler is very long (need bypass pythonize)
* When scheduler unserialize, uuid is not defined and next what_i_managed is asked by arbiter and not have finished unserialize, it send none and arbiter resend conf (infinite loop)
* Start all daemons. after couple of minutes, stop broker. Start broker. Arbiter ask what_i_manage and broker send nothing but arbiter never try to send the config to broker
* $PLUGINSDIR$ replaced with '' instead the value in config

alignak/alignakobject.py
```diff
@@ -32,13 +32,26 @@ class AlignakObject(object):
...
     def __init__(self, params=None, parsing=True):  # pylint: disable=W0613
 
         if params is None:
             return
+        hasmacro = False
+        if hasattr(self, 'macros'):
+            hasmacro = True
         for key, value in params.iteritems():
-            setattr(self, key, value)
+            if key in ['already_start_escalations', 'tags', 'notified_contacts',
```
Seb-Solon: You should not be based on KEY in AlignakObject. If you want sets just test if key in self.properties and is a SetProp. The list here is only for a few objects.

alignak/alignakobject.py
```diff
@@ -32,13 +32,26 @@ class AlignakObject(object):
...
+                       'parent_dependencies', 'child_dependencies']:
+                setattr(self, key, set(value))
+            else:
+                setattr(self, key, value)
+            # reconstruct macro list
+            if hasmacro:
```
Seb-Solon: Why do we need this case? We should already have params well formatted and the previous loop should do the job.

alignak/brok.py
```diff
@@ -53,19 +53,26 @@
...
 from alignak.misc.serialization import serialize, unserialize, AlignakClassLookupException
-from alignak.alignakobject import AlignakObject
 
 
-class Brok(AlignakObject):
+class Brok(object):
     """A Brok is a piece of information exported by Alignak to the Broker.
     Broker can do whatever he wants with it.
     """
-    __slots__ = ('__dict__', 'uuid', 'type', 'data', 'prepared', 'instance_id')
```
Seb-Solon: I think we should keep this for now

alignak/brok.py
```diff
@@ -53,19 +53,26 @@
...
     my_type = 'brok'
 
     def __init__(self, params, parsing=True):
         if not parsing:
-            super(Brok, self).__init__(params, parsing)
+            if params is None:
+                return
+            for key, value in params.iteritems():
+                if key in ['already_start_escalations', 'tags', 'notified_contacts',
```
Seb-Solon: See my previous comment.

alignak/http/arbiter_interface.py
```diff
@@ -62,16 +62,6 @@ def put_conf(self, conf):
 
     @cherrypy.expose
     @cherrypy.tools.json_out()
-    def get_config(self):
```
Seb-Solon: was not used?

alignak/http/generic_interface.py
```diff
@@ -224,9 +222,12 @@ def get_external_commands(self):
         :return: serialized external command list
         :rtype: str
         """
-        with self.app.external_commands_lock:
-            cmds = self.app.get_external_commands()
-            raw = serialize(cmds)
+        if hasattr(self.app, 'external_commands_lock'):
```
Seb-Solon: It shoud be always true. If not then add it. The lock ensure we don't loose external commands. Maybe the if is on the get_external_commands method. How did you hit this case?

alignak/http/generic_interface.py
```diff
@@ -224,9 +222,12 @@ def get_external_commands(self):
         :return: serialized external command list
         :rtype: str
```
Seb-Solon: type does not match any more

alignak/http/generic_interface.py
```diff
@@ -269,7 +270,8 @@ def get_broks(self, bname):  # pylint: disable=W0613
         """
         with self.app.lock:
             res = self.app.get_broks()
-            return serialize(res, True)
+
+        return serialize(res, True)
```
Seb-Solon: Why do we need that?

alignak/http/generic_interface.py
```diff
@@ -282,17 +284,18 @@ def get_raw_stats(self):
...
-            for mod in app.q_by_mod:
-                # In workers we've got actions send to queue - queue size
-                for (q_id, queue) in app.q_by_mod[mod].items():
-                    lst.append({
-                        'scheduler_name': sched['name'],
-                        'module': mod,
-                        'queue_number': q_id,
-                        'queue_size': queue.qsize(),
-                        'return_queue_len': app.get_returns_queue_len()})
+        if hasattr(app, 'schedulers'):
```
Seb-Solon: When it is possible to no have it?

alignak/objects/item.py
```diff
@@ -155,6 +155,7 @@ def __init__(self, params=None, parsing=True):
                               "it is not in %s object properties" % \
                               (key, cls.__name__)
                     self.configuration_warnings.append(warning)
+                    self.properties[key] = ToGuessProp(default='')
```
Seb-Solon: Why is it necessary?

test/full_tst.py
```diff
@@ -22,10 +22,8 @@
 import subprocess
 import json
 from time import sleep
-import urllib
+import requests
```
Seb-Solon: \o/

alignak/http/daemon.py
```diff
@@ -95,7 +95,8 @@ def __init__(self, host, port, http_interface, use_ssl, ca_cert,
 
         self.srv = CherryPyWSGIServer((host, port),
                                       cherrypy.Application(http_interface, "/", config),
-                                      numthreads=daemon_thread_pool_size, shutdown_timeout=1)
+                                      numthreads=daemon_thread_pool_size, shutdown_timeout=1,
+                                      request_queue_size=30)
```
Seb-Solon: what it is for?

alignak/alignakobject.py
ddurieux: Humm, yes, you're right, will update this

alignak/brok.py
ddurieux: With this, we have errors, without it works like a charm

alignak/http/arbiter_interface.py
ddurieux: No call to this API resource in all alignak code, so I have deleted it

alignak/http/generic_interface.py
ddurieux: One of the daemons (not sure witch not have this and so prevent error in cherrypy)

alignak/http/generic_interface.py
ddurieux: I think it's better to not lock when serialize

alignak/http/generic_interface.py
ddurieux: For one or two (not keep in my memory) daemons, this attribute not exist, so prevent 500 error in cherrypy

alignak/objects/item.py
ddurieux: In the case property not defined type (like custom properties), if we don't set it, it make an error

test/full_tst.py
ddurieux: Yes it's better to test with same lib in code of alignak ^^

alignak/http/daemon.py
ddurieux: It modify queue length of cherrypy socket server from 5 to 30 (better for big config)

alignak/objects/item.py
Seb-Solon: Where the error occurs?

ddurieux: The properties defined in modules (additional modules), are not pass after serialization because they not have property type (so it's guessprop) and the module not have these properties if this line not added

ddurieux: We need to add it to propertie dict in order to get this property serialized later on

PR #298, opened 2016-05-09
Replace thread.rlock by queue.Queue for consume result in scheduler. …

### LINKED ISSUE #297: Scheduler put_results and thread rlock
Actually, the endpoint _put_results_ of scheduler receive the result and put them in _self.app.sched.waiting_results_ with a thread lock

The scheduler process, copy the _waiting_results_ and define empty in the 'same' thread lock.

Why not prefer use Queue and so _put()_ and _get()_ functions where it's the goal we want instead re-code a _home_ queue not really optimized?

ddurieux:
…closes #297

### DIFF
--- a/alignak/scheduler.py
+++ b/alignak/scheduler.py
@@ -72,7 +72,7 @@ import os
 import cStringIO
 import tempfile
 import traceback
-import threading
+from Queue import Queue
 from collections import defaultdict
 
 from alignak.external_command import ExternalCommand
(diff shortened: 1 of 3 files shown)


alignak/scheduler.py
```diff
@@ -195,8 +194,10 @@ def reset(self):
         :return: None
         """
         self.must_run = True
-        with self.waiting_results_lock:
-            del self.waiting_results[:]
+
+        queue_size = self.waiting_results.qsize()
+        for _ in xrange(queue_size):
+            _ = self.waiting_results.get()
```
Seb-Solon: Have a look at this to empty queue : http://stackoverflow.com/questions/6517953/clear-all-items-from-the-queue

PR #310, opened 2016-06-28
Rewrite all unit tests + fix bugs found

ddurieux:
Rewrite unittest and pass them in pep8 & pylint
So fix too the bugs found when rework tests


.travis.yml
```diff
@@ -21,7 +21,7 @@ script:
   - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then find -iname '*.pyc' -exec rm -rf {} \; && travis_wait pylint --rcfile=.pylintrc -r no alignak; fi
   - export PYTHONWARNINGS=all
   - pep257 --select=D300 alignak
-  - cd test && (pkill -6 -f "alignak_-" || :) && python full_tst.py && cd ..
+  - cd test && (pkill -6 -f "alignak_-" || :) && sudo mkdir /var/run/alignak && sudo chmod 755 alignak /var/run/alignak && python full_tst.py && cd ..
```
Seb-Solon: the idea is to drop sudo in travis in order to use their new infrastructure. So better use the config file inside test/etc

ddurieux: Ok, I have another idea, will be better ;)

alignak/objects/schedulingitem.py
```diff
@@ -119,6 +119,8 @@ class SchedulingItem(Item):  # pylint: disable=R0902
             BoolProp(default=False, fill_brok=['full_status']),
         'freshness_threshold':
             IntegerProp(default=0, fill_brok=['full_status']),
+        'freshness_state':
```
Seb-Solon: It belong to class Service not SchedulingItem

ddurieux: It can be here, I don't see really difference

alignak/scheduler.py
```diff
@@ -119,16 +119,15 @@ def __init__(self, scheduler_daemon):
         self.recurrent_works = {
             0: ('update_downtimes_and_comments', self.update_downtimes_and_comments, 1),
             1: ('schedule', self.schedule, 1),  # just schedule
-            2: ('consume_results', self.consume_results, 1),  # incorporate checks and dependencies
+            2: ('check_freshness', self.check_freshness, 10),
```
Seb-Solon: Why change check freshness order? To get the check freshness and get the result in the same loop? The poller may take some time to do it

alignak/scheduler.py
```diff
@@ -1331,7 +1338,7 @@ def get_retention_data(self):
                         if fun:
                             val = fun(serv, val)
                         s_dict[prop] = val
-            all_data['services'][(serv.host.host_name, serv.service_description)] = s_dict
+            all_data['services'][(serv.host_name, serv.service_description)] = s_dict
```
Seb-Solon: nice catch

alignak/scheduler.py
```diff
@@ -1774,16 +1782,20 @@ def get_new_broks(self):
...
 
         :return: None
         """
-        # print "********** Check freshness******"
         for elt in self.iter_hosts_and_services():
-            chk = elt.do_check_freshness(self.hosts, self.services, self.timeperiods,
-                                         self.macromodulations, self.checkmodulations, self.checks)
-            if chk is not None:
-                self.add(chk)
+            if elt.check_freshness and elt.passive_checks_enabled:
```
Seb-Solon: This if is already included in the function itself.

alignak/scheduler.py
ddurieux: The poller? but freshness is for passive, so we create check result, so we have the result immediatly, it's more logical like this :p

alignak/scheduler.py
ddurieux: Yes but it keep some if and code if not required, so in case you have many active checks, we keep cpu ;)

alignak/scheduler.py
Seb-Solon: Yes but the check freshness can trigger a forced check that is executed by a poller

alignak/scheduler.py
Seb-Solon: what are you keeping? Stack write? We are doing double check here. If we want to change the behavior we will need to this it in both place

alignak/scheduler.py
ddurieux: No, we have discussed this with mohierf and if it's a only passive, not have _active_ check to run, so I have deleted this possibility (if it's only passive, never run the check_command). 

On change status the eventhandler can work ;)

And like before, there is not very different like this, because the consume_result will be called next loop (so less than 1 second) and so not very logical

alignak/scheduler.py
ddurieux: You're right, will remove the check in the function elt.do_check_freshness

test/alignak_test.py
```diff
@@ -136,102 +137,92 @@ class AlignakTest(unittest.TestCase):
...
-        self.external_command_dispatcher = e2
-        self.sched.conf.accept_passive_unknown_check_results = False
-
-        self.sched.schedule()
+    @classmethod
+    def setUp(cls):
+        """
+        Load default configuration when start test
+
+        @verified
```
Seb-Solon: ?

test/alignak_test.py
```diff
@@ -136,102 +137,92 @@ class AlignakTest(unittest.TestCase):
...
-        self.sched.external_command = e
-        e.load_scheduler(self.sched)
-        e2 = ExternalCommandManager(self.conf, 'dispatcher')
-        e2.load_arbiter(self)
-        self.external_command_dispatcher = e2
-        self.sched.conf.accept_passive_unknown_check_results = False
-
-        self.sched.schedule()
+    @classmethod
+    def setUp(cls):
```
Seb-Solon: please use self

test/alignak_test.py
```diff
@@ -136,102 +137,92 @@ class AlignakTest(unittest.TestCase):
...
+                        sched.accept_passive_unknown_check_results,
+                    'api_key': cls.arbiter.dispatcher.conf.api_key,
+                    'secret': cls.arbiter.dispatcher.conf.secret,
+                    'http_proxy': cls.arbiter.dispatcher.conf.http_proxy,
+                    'statsd_host': cls.arbiter.dispatcher.conf.statsd_host,
+                    'statsd_port': cls.arbiter.dispatcher.conf.statsd_port,
+                    'statsd_prefix': cls.arbiter.dispatcher.conf.statsd_prefix,
+                    'statsd_enabled': cls.arbiter.dispatcher.conf.statsd_enabled,
+                }
+                cls.scheduler.new_conf = conf_package
```
Seb-Solon: You are overriding the same variable. Only the last loop will sork (I guess there is only one elem in loop)

test/alignak_test.py
ddurieux: I modify functions in alignak_test.py to be ok and it's annotation for the mment to see if the function is ok or not. At the end I will delete this line

test/alignak_test.py
ddurieux: You're right

test/alignak_test.py
ddurieux: Yes there is only one

test/alignak_test.py
```diff
@@ -136,102 +137,92 @@ class AlignakTest(unittest.TestCase):
...
+                cls.scheduler.setup_new_conf()
+
+    def setup_with_file(self, configuration_file):
```
Seb-Solon: what is the difference with setup?

alignak/dispatcher.py
```diff
@@ -373,283 +366,268 @@ def get_scheduler_ordered_list(self, realm):
...
+            logger.info("OK, all schedulers configurations are dispatched :)")
+
+        # Sched without conf in a dispatch ok are set to no need_conf
+        # so they do not raise dispatch where no use
+        for sched in self.schedulers.items.values():
+            if sched.conf is None:
+                # "so it do not ask anymore for conf"
+                sched.need_conf = False
+
+    def prepare_dispatch_other_modules(self, sat_type, realm, cfg, arbiters_cfg):
```
Seb-Solon: dispatch other satellite maybe?

alignak/daemons/receiverdaemon.py
```diff
@@ -279,6 +276,10 @@ def setup_new_conf(self):
                 self.have_modules = True
                 logger.info("We received modules %s ", mods)
 
+                self.do_load_modules(self.modules)
```
Seb-Solon: any reason to put this upper?

alignak/dispatcher.py
```diff
@@ -224,8 +221,15 @@ def check_dispatch(self):  # pylint:disable=R0912
                         sched.push_flavor = 0
                         sched.need_conf = True
                         sched.conf = None
-                    # Else: ok the conf is managed by a living scheduler
 
+        self.check_disptach_other_modules()
+
+    def check_disptach_other_modules(self):
```
Seb-Solon: satellite?

alignak/objects/config.py
```diff
@@ -2467,9 +2443,12 @@ def create_packs(self, nb_packs):  # pylint: disable=R0915,R0914,R0912,W0613
                 if valid_value and old_pack in packindices:
                     # print 'Use a old id for pack', old_pack, [h.get_name() for h in pack]
                     i = old_pack
-                else:  # take a new one
-                    # print 'take a new id for pack', [h.get_name() for h in pack]
-                    i = round_robin.next()
+                else:
+                    if isinstance(i, int):
```
Seb-Solon: only true with -1 right?

alignak/scheduler.py
ddurieux: I have removed duplicated code/checks in function _do_check_freshness_

alignak/daemons/receiverdaemon.py
ddurieux: If I remember, have errors

alignak/objects/config.py
ddurieux: Yep

alignak/objects/config.py
```diff
@@ -1099,13 +1075,14 @@ def read_config_buf(self, buf):  # pylint: disable=R0912
 
             line_nb += 1
             # Remove comments
-            line = split_semicolon(line)[0].strip()
+            if not continuation_line:
```
Seb-Solon: What is this fixing?

alignak/objects/config.py
```diff
@@ -1099,13 +1075,14 @@ def read_config_buf(self, buf):  # pylint: disable=R0912
...
+                line = split_semicolon(line)[0].strip()
 
             # A backslash means, there is more to come
             if re.search(r"\\\s*$", line) is not None:
                 continuation_line = True
                 line = re.sub(r"\\\s*$", "", line)
-                line = re.sub(r"^\s+", " ", line)
+                # line = re.sub(r"^\s+", " ", line)
```
Seb-Solon: What is this fixing?

alignak/objects/config.py
```diff
@@ -2146,9 +2132,19 @@ def is_correct(self):  # pylint: disable=R0912
...
+                if self.read_config_silent == 0:
+                    logger.info('Checked %s, not correct!', obj)
+
                 valid = False
-                logger.error("\t%s conf incorrect!!", obj)
+                self.configuration_errors += cur.configuration_errors
+                logger.error("\t%s configuration is incorrect!", obj)
+            if cur.configuration_warnings:
+                self.configuration_warnings += cur.configuration_warnings
+                logger.error("\t%s configuration warnings: %d, total: %d", obj,
```
Seb-Solon: should be logger.warning

alignak/objects/config.py
```diff
@@ -2132,8 +2109,17 @@ def is_correct(self):  # pylint: disable=R0912
 
             cur = getattr(self, obj)
             if not cur.is_correct():
```
Seb-Solon: We can refactor those 2 block I think

alignak/objects/host.py
```diff
@@ -518,11 +524,14 @@ def set_state_from_exit_status(self, status, notif_period, hosts, services):
...
             state_code = 'd'
+
         if state_code in self.flap_detection_options:
             self.add_flapping_change(self.state != self.last_state)
             # Now we add a value, we update the is_flapping prop
             self.update_flapping(notif_period, hosts, services)
-        if self.state != self.last_state:
+
+        if self.state != self.last_state and \
+                not (self.state == "DOWN" and self.last_state == "UNREACHABLE"):
```
Seb-Solon: What it is fixing?

alignak/daemons/arbiterdaemon.py
```diff
@@ -664,8 +661,7 @@ def run(self):
         # Before running, I must be sure who am I
         # The arbiters change, so we must re-discover the new self.me
         for arb in self.conf.arbiters:
-            print "ARR3:", arb
-            if arb.is_me():
+            if arb.get_name() == self.config_name:
```
mohierf: You are right! I will change this

alignak/objects/config.py
mohierf: Why not ... I left ERROR to be more alerting. I will change this

alignak/objects/config.py
ddurieux: Because need get spaces and all info on multiline command for exemple when have html code, see https://github.com/Alignak-monitoring/alignak/blob/develop%2Benhanceunittests/test/test_command_multilines.py

alignak/objects/config.py
ddurieux: Same than previously

alignak/objects/config.py
mohierf: The two for...loop may be grouped as one loop, yes!

alignak/objects/config.py
```diff
@@ -1068,6 +1068,7 @@ def read_config_buf(self, buf):  # pylint: disable=R0912
         lines = buf.split('\n')
         line_nb = 0  # Keep the line number for the file path
         for line in lines:
+            filefrom = ''
```
mohierf: I will revert this modification because it breaks some tests in test_config. The importing filename is missing ...

ddurieux:
I close

PR #312, opened 2016-07-11
Fix load modules in scheduler

ddurieux:


alignak/daemons/schedulerdaemon.py
```diff
@@ -254,7 +254,7 @@ def setup_new_conf(self):
 
             self.cur_conf = conf
             self.override_conf = override_conf
-            self.modules = modules
+            self.modules = unserialize(modules, True)
```
Seb-Solon: Maybe we should unserialize new_c = self.new_conf then

Seb-Solon:
The fix seems legit, but it need a test still :(

ddurieux:
Replaced in branch develop+enhanceunittests (see commit above)

PR #319, opened 2016-07-25
Enh: Remove shinken ref in etc

Seb-Solon:


etc/alignak.cfg
```diff
@@ -124,14 +124,6 @@ use_ssl=0
 #hard_ssl_name_check=0
 
 
-# kernel.alignak.io communication channel. Create an account to http://shinken.io
-# and look at your profile to fill this.
-#api_key=
-#secret=
-# if you need an http proxy to exchange with kernel.alignak.io
-#http_proxy=
```
ddurieux: Perhaps we can keep _api_key_, _secret_ and _http_proxy_ because we use it for statsd

Seb-Solon:
statsd pushes to "api.alignak.io" and it is not configurable. For not I am jsut hiding non usable parameters

mohierf:
It should be interesting to include all the possible parameters with their default values? Such as:
```
# Log configuration
# Notifications
# log_notifications=1

# Services retries
# log_service_retries=1

# Hosts retries
# log_host_retries=1

# Event handlers
# log_event_handlers=1

# External commands
# log_external_commands=1

# Passive checks
# log_passive_checks=1

# Initial states
# log_initial_states=1
```

Seb-Solon:
It was not really the main target of the Pull request...

PR #331, opened 2016-09-12
WiP: Clean elements groups and refactor configuration warnings/errors logging

### LINKED ISSUE #93: Hostgroup does not store its hostgroups_members
See https://github.com/naparuba/shinken/issues/1652

### LINKED ISSUE #110: Add property hostgroup_members in Hostgroup class
When comment modules (pylint), I have seen the function get_hostgroup_members of Hostgroup never works because property 'hostgroup_members' not exist.

So clean it and remove the call in function 'get_hosts_by_explosion' in same class

mohierf:
- replace groups members property from String to List, this to avoid lists with one empty string element
- fixes #93 
- fixes #110 
- refactor unit tests for groups
- refactor configuration warnings/errors logging


alignak/objects/config.py
```diff
@@ -2234,12 +2256,24 @@ def add_error(self, txt):
...
+        on the end of configuration parsing when an error has been detected.
 
         :return:  None
         """
-        for err in self.configuration_errors:
-            logger.error(err)
+        if self.configuration_warnings:
+            logger.info("Configuration warnings:")
+            for msg in self.configuration_warnings:
+                logger.info(msg)
```
ddurieux: for warning and errors, use _logger.warning_ instead _logger.info_

mohierf: No, bad idea. As explained in the comment, using INFO is better beceause these logs are simply used to build a sum-up when configuration check is finished. If we use WARNING and ERROR, each warning and error will appera twice in the logs ... bad idea if we want to parse the logs with an application like LogStash :/

alignak/objects/service.py
```diff
@@ -301,6 +301,14 @@ def get_name(self):
             return self.name
         return 'SERVICE-DESCRIPTION-MISSING'
 
+    def get_servicegroups(self):
```
ddurieux: We can remove this function by replace _get_servicegroups()_ calls by self.servicegroups

mohierf: Yes, why not ... but there is so much functions like this one. We should organize another code cleaning session to fix all this kind of functions later.

ddurieux: okay

PR #342, opened 2016-09-16
Add more tests made in a separate  branch

mohierf:


alignak/modulesmanager.py
```diff
@@ -93,10 +93,10 @@ def set_max_queue_size(self, max_queue_size):
     def load_and_init(self, mod_confs):
         """Import, instantiate & "init" the modules we have been requested
 
-        :return: None
+        :return: Number of module instances loaded and initialized
```
ddurieux: add after: 
:rtype: int

PR #348, opened 2016-09-19
Move all old tests to _old + rewrite alignak_test.py to use scheduler…

ddurieux:
…, arbiter... satellites with more nearly behavior than with old tests

it's first PR about rewrite test and so fix bugs found in same time ;)

* all tests moved in _old
* rewrite test_alignak.py (no need check review it)


test/alignak_test.py
```diff
@@ -131,127 +132,94 @@ class Pluginconf(object):
...
+                sched.setup_new_conf()
+            self.schedulers.append(sched)
 
     def add(self, b):
         if isinstance(b, Brok):
             self.broks[b.uuid] = b
             return
         if isinstance(b, ExternalCommand):
-            self.sched.run_external_command(b.cmd_line)
+            self.schedulers[0].run_external_command(b.cmd_line)
```
Seb-Solon: So we add all broker to the first sched? Maybe we could at least always take the "scheduler-master" or something. So that we ensure we add to the same scheduler. This may required to have schedulers as dict and not lists

test/alignak_test.py
```diff
@@ -271,10 +239,95 @@ def fake_check(self, ref, exit_status, output="OK"):
         check.exit_status = exit_status
         check.execution_time = 0.001
         check.status = 'waitconsume'
-        self.sched.waiting_results.put(check)
-
-    def scheduler_loop(self, count, reflist, do_sleep=False, sleep_time=61, verbose=True,
-                       nointernal=False):
+        self.schedulers[0].sched.waiting_results.put(check)
```
Seb-Solon: Same for all schedulers[0]

test/alignak_test.py
```diff
@@ -271,10 +239,95 @@ def fake_check(self, ref, exit_status, output="OK"):
...
+                    for i in self.schedulers[0].sched.recurrent_works:
+                        (name, fun, nb_ticks) = self.schedulers[0].sched.recurrent_works[i]
+                        if nb_ticks == 1:
+                            fun()
+                self.assertGreater(len(obj.checks_in_progress), 0)
+                chk = self.schedulers[0].sched.checks[obj.checks_in_progress[0]]
+                chk.set_type_active()
+                chk.output = output
+                chk.exit_status = exit_status
+                self.schedulers[0].sched.waiting_results.put(chk)
```
Seb-Solon: can you add a bit more comment? From what I get : we create new check by running recurrent work if necessary. Then you manully do the launch check part and make check to get consumed (recurrent works)

test/alignak_test.py
```diff
@@ -286,180 +339,384 @@ def scheduler_loop(self, count, reflist, do_sleep=False, sleep_time=61, verbose=
...
-        if hasattr(self, "sched"):
-            actions = self.sched.actions
-        else:
-            actions = self.actions
-        for a in sorted(actions.values(), lambda x, y: cmp(x.uuid, y.uuid)):
+        actions = sorted(self.schedulers[0].sched.actions.values(), key=lambda x: x.creation_time)
+        for a in actions:
             if a.is_a == 'notification':
-                item = self.sched.find_item_by_id(a.ref)
+                item = self.scheduler.sched.find_item_by_id(a.ref)
```
Seb-Solon: schedulers[0] ??

test/alignak_test.py
```diff
@@ -286,180 +339,384 @@ def scheduler_loop(self, count, reflist, do_sleep=False, sleep_time=61, verbose=
...
                 if item.my_type == "host":
                     ref = "host: %s" % item.get_name()
                 else:
-                    hst = self.sched.find_item_by_id(item.host)
+                    hst = self.scheduler.sched.find_item_by_id(item.host)
```
Seb-Solon: schedulers[0] ??

test/alignak_test.py
```diff
@@ -286,180 +339,384 @@ def scheduler_loop(self, count, reflist, do_sleep=False, sleep_time=61, verbose=
...
+        @verified
+
+        :param number: number of actions we must have
+        :type number: int
+        :return: None
+        """
+        print("Actions: %s" % self.schedulers[0].sched.actions)
+        actions = sorted(self.schedulers[0].sched.actions.values(), key=lambda x: x.creation_time)
+        self.assertEqual(number, len(self.schedulers[0].sched.actions),
+                         "Not found expected number of actions:\nactions_logs=[[[\n%s\n]]]" %
```
Seb-Solon: brainfuck spotted

test/alignak_test.py
```diff
@@ -131,127 +132,94 @@ class Pluginconf(object):
 class AlignakTest(unittest.TestCase):
 
     time_hacker = TimeHacker()
+    maxDiff = None
```
Seb-Solon: What is that for?

ddurieux: It's uses by nosetests to display a complete diff between 2 dicts in case self.assertEqual() not ok

alignak/objects/satellitelink.py
```diff
@@ -241,23 +240,24 @@ def add_failed_check_attempt(self, reason=''):
...
-        since_last_check = now - self.last_check
-        if since_last_check < self.check_interval:
+        if (now - self.last_check) < self.check_interval:
             return
 
         self.last_check = now
 
         # We ping and update the managed list
         self.ping()
+        if not self.alive or self.attempt > 0:
```
Seb-Solon: By skipping the update call self.managed_confs may have old data. Is is an issue? (I did not check maybe you did already)

alignak/util.py
```diff
@@ -774,6 +774,24 @@ def sort_by_ids(x00, y00):
     return 0
 
 
+def sort_by_number_values(x00, y00):
+    """Compare x00, y00 base on number of values
+
+    :param x00: first elem to compare
+    :type x00: int
```
Seb-Solon: int has no len parameter. I guess type are not good.

test/test_dispatcher.py
```diff
@@ -0,0 +1,448 @@
...
+        * 1 receiver
+        * 1 reactionner
+        * 1 broker
+
+        :return: None
+        """
+        self.setup_with_file('cfg/cfg_dispatcher_simple_multi_schedulers.cfg')
+        self.assertEqual(1, len(self.arbiter.dispatcher.realms))
+        for realm in self.arbiter.dispatcher.realms:
+            self.assertEqual(2, len(realm.confs))
```
Seb-Solon: Hum if you have 2 realm then you shuold have 2 broker. As far as I remember a realm is at least 1 sched and 1 brok. Unless this is a sub realm maybe?

test/test_dispatcher.py
```diff
@@ -0,0 +1,448 @@
...
+
+        self.assertSetEqual(set([4, 6]), set([len(self.schedulers['scheduler-master'].sched.hosts),
+                                              len(self.schedulers['realm2scheduler-master'].sched.hosts)]))
+
+        # for satellite in self.arbiter.dispatcher.satellites:
+        #     self.assertNotEqual({}, satellite.cfg['schedulers'], satellite.get_name())
+        #     self.assertEqual(1, len(satellite.cfg['schedulers']),
+        #                      'must have 1 scheduler in {0}'.format(satellite.get_name()))
+
+    def test_realms_with_sub(self):
```
Seb-Solon: I pretty sure we drifted with the initial definition of a realm here.

test/test_dispatcher.py
```diff
@@ -0,0 +1,448 @@
...
+        pass
+
+    def test_simple_scheduler_spare(self):
+        """
+        Test simple but with spare of scheduler
+
+        :return: None
+        """
+        with requests_mock.mock() as mockreq:
+            mockreq.get('http://localhost:7768/ping', json='pong')
```
Seb-Solon: for loop?

test/test_dispatcher.py
```diff
@@ -0,0 +1,448 @@
...
+            mockreq.get('http://localhost:7772/ping', json='pong')
+            mockreq.get('http://localhost:7771/ping', json='pong')
+            mockreq.get('http://localhost:7769/ping', json='pong')
+            mockreq.get('http://localhost:7773/ping', json='pong')
+            mockreq.get('http://localhost:8002/ping', json='pong')
+
+            self.setup_with_file('cfg/cfg_dispatcher_scheduler_spare.cfg')
+            json_managed = {self.schedulers['scheduler-master'].conf.uuid:
+                            self.schedulers['scheduler-master'].conf.push_flavor}
+            mockreq.get('http://localhost:7768/what_i_managed', json=json_managed)
```
Seb-Solon: for loop?

test/test_dispatcher.py
```diff
@@ -0,0 +1,448 @@
...
+        for satellite in self.arbiter.dispatcher.reactionners:
+            satellite.check_interval = 1
+        for satellite in self.arbiter.dispatcher.brokers:
+            satellite.check_interval = 1
+        for satellite in self.arbiter.dispatcher.pollers:
+            satellite.check_interval = 1
+        time.sleep(1)
+
+        with requests_mock.mock() as mockreq:
+            mockreq.get('http://localhost:7772/ping', json='pong')
```
Seb-Solon: for loop?

test/test_dispatcher.py
```diff
@@ -0,0 +1,448 @@
...
+            for satellite in self.arbiter.dispatcher.satellites:
+                self.assertEqual(1, len(satellite.cfg['schedulers']))
+                scheduler = satellite.cfg['schedulers'].itervalues().next()
+                self.assertEqual('scheduler-spare', scheduler['name'])
+                json_managed_spare[scheduler['instance_id']] = scheduler['push_flavor']
+
+        # return of the scheduler master
+        print "*********** Return of the king / master ***********"
+        with requests_mock.mock() as mockreq:
+            mockreq.get('http://localhost:7768/ping', json='pong')
```
Seb-Solon: for loop?

alignak/daemons/arbiterdaemon.py
```diff
@@ -664,8 +665,7 @@ def run(self):
         # Before running, I must be sure who am I
         # The arbiters change, so we must re-discover the new self.me
         for arb in self.conf.arbiters:
-            print "ARR3:", arb
-            if arb.is_me():
+            if arb.get_name() == self.config_name:
```
ddurieux: You're right, I modify it

alignak/util.py
```diff
@@ -1308,6 +1308,10 @@ def parse_daemon_args(arbiter=False):
                                  'multiple -c can be used, they will be concatenated')
         parser.add_argument("-V", "--verify-config", dest="verify_only", action="store_true",
                             help="Verify config file and exit")
+        parser.add_argument("-n", "--config-name", dest="config_name",
```
ddurieux: We pass the arbiter name when start, will be useful for spare feature (when we will code it).
The second problem with the spare it's each arbiter (master and spare) try get name with is_me() but can't really use it because not get really arbiter name.

alignak/util.py
ddurieux: Yes it's list, I fix it

alignak/objects/satellitelink.py
```diff
@@ -91,9 +89,10 @@ class SatelliteLink(Item):
         'attempt':              StringProp(default=0, fill_brok=['full_status']),
 
         # can be network ask or not (dead or check in timeout or error)
-        'reachable':            BoolProp(default=False, fill_brok=['full_status']),
-        'last_check':           IntegerProp(default=0, fill_brok=['full_status']),
-        'managed_confs':        StringProp(default={}),
+        'reachable': BoolProp(default=True, fill_brok=['full_status']),
```
ddurieux: Because in dispatcher, I modified the behavior.

By default we prepare conf to a satellite (scheduler, poller...)

After we send, it is reachable by default, so we begin attempts, if not ok at xx attempts, we prepare for another satellite (if we have many) or we need wait satellite be up

alignak/objects/satellitelink.py
```diff
@@ -241,23 +240,24 @@ def add_failed_check_attempt(self, reason=''):
         if self.attempt == self.max_check_attempts:
             self.set_dead()
 
-    def update_infos(self):
+    def update_infos(self, now):
```
ddurieux: Only because we reduce now() calls in dispatcher.check_alive() because can have many satellites.
Seems better use it like this no?

alignak/objects/satellitelink.py
ddurieux: Not need to run the end of the function (check conf managed) if satellite not available because we will try connection again for nothing because 0.2 second before it says not available

test/test_dispatcher.py
ddurieux: A broker can be subrealm ;)

test/test_dispatcher.py
ddurieux: ah? you are sure?

test/test_dispatcher.py
ddurieux: With that, I can simulate answer or not the satellites ;)

ddurieux: Ok you're right, I update it

test/test_dispatcher.py
ddurieux: I update it

test/test_dispatcher.py
ddurieux: I update it

test/test_dispatcher.py
ddurieux: I update it

PR #355, opened 2016-09-22
WiP: Setup and default configuration

mohierf:
Update setup.py script:
- check if an `alignak` user exists before installing
- update *resource.d/paths.cfg* file after installation
- update configuration files with or without comments according to variable type

Update default configuration:
- renamed some directories (*arbiter_cfg* and *daemons_cfg*)
- removed sample and useless files
- moved *templates* and *packs* upper
- cleaned and commented all files
- add a *packs/resource.d* directory for the global macros installed byt the packs
- make the daemons use current logged-in user as default


mohierf:
Forget this PR!

PR #364, opened 2016-09-25
Replace localhost address by 127.0.0.1 to prevent problem on linux di…

ddurieux:
…stributions have ipv4 and ipv6 activated


Seb-Solon:
+ 1 we dont have ipv6 support tested

PR #383, opened 2016-09-30
WiP: Develop+dependencies

### LINKED ISSUE #382: Service dependencies for a simple hostgroup
**EDIT**: there is no real bug in this issue but a quite tricky problem with some standard `servicedependency` definition.

-----

**1/**
When defining a `servicedependency` as is:
```
define servicedependency {
    hostgroup_name nrpe_hosts
    service_description NRPE Status
    dependent_hostgroup_name nrpe_hosts
    dependent_service_description CPU Load,Disk Usage
    }
```

one expect that:
- Any host that is a member of the hostgroup will get the service dependency
- The services *CPU Load* and *Disk Usage* depends upon the service *NRPE Status* on all hosts of the hostgroup
- The hosts in the hostgroup **DO HAVE**  a relationship with each other

**2/**
When defining a `servicedependency` as is:
```
define servicedependency {
    hostgroup_name nrpe_hosts
    service_description NRPE Status
    dependent_service_description CPU Load,Disk Usage
    }
```

one expect that:
- Any host that is a member of the hostgroup will get the service dependency
- The services *CPU Load* and *Disk Usage* depends upon the service *NRPE Status* on all hosts of the hostgroup
- The hosts in the hostgroup **DO NOT HAVE** relationship with each other

---
In both case, Alignak creates a relation between each host of the hostgroup. 

Let us assume that we have 2 hosts in the hostgroup, we expect to have only 4 dependencies with the first definition and 8 with the second one, but we always have 8 dependencies ... because:
- host1/CPU Load depends on host1/NRPE Status
- host1/Disk Usage depends on host1/NRPE Status
- host2/CPU Load depends on host2/NRPE Status
- host2/Disk Usage depends on host2/NRPE Status

**and**

- host1/CPU Load depends on host**2**/NRPE Status
- host1/Disk Usage depends on host**2**/NRPE Status
- host2/CPU Load depends on host**1**/NRPE Status
- host2/Disk Usage depends on host**1**/NRPE Status

The 4 last dependencies should not be defined in the first case.

mohierf:
Fixes #382 : dispatch service dependencies to all hosts of an hostgroup

Update unit tests for hosts and services dependencies

----
**Note**: this PR will be adapted after the current huge testing work made by @ddurieux


mohierf:
The Travis build fails because of tests ... waiting for some modifications done elsewhere :wink:

mohierf:
Close this PR to rebuild on the new develop branch ... too much modifications to manage!

PR #385, opened 2016-10-02
Improve configuration load.

ddurieux:


alignak/objects/service.py
```diff
@@ -1589,7 +1593,8 @@ def explode(self, hosts, hostgroups, contactgroups,
 
         # Then for every host create a copy of the service with just the host
         # because we are adding services, we can't just loop in it
-        for s_id in self.items.keys():
+        itemkeys = self.items.keys()
```
Seb-Solon: What is the advantage of doing that?

ddurieux: I have pylint error on use _for s_id in self.items.keys()_ because it say we can use _for s_id in self.items_ but in this loop we can delete the key, so it's the only way I found to not have error and not disable the pylint on this

PR #386, opened 2016-10-02
Rework logs

Seb-Solon:
Use logging lib to have parent logger instaed of using the same logger constant.

This is the first step of logging rework


test/alignak_test.py
```diff
@@ -162,12 +175,30 @@ def setup_with_file(self, configuration_file):
...
         try:
+            # The following is copy paste from setup_alignak_logger
+            # The only difference is that keep logger at INFO level to gather messages
+            # This is needed to assert later on logs we received.
+            self.logger.setLevel('INFO')
+            # Force the debug level if the daemon is said to start with such level
+            if self.arbiter.debug:
+                self.logger.setLevel('DEBUG')
+
+            # Log will be broks
```
mohierf: ? Why this comment ?

Seb-Solon: Yeah copy paste from setup logger function

Seb-Solon: But it is still true. Those logs are broks with the BrokHandler in the logger

PR #390, opened 2016-10-03
Add test for acknowledge and fix bugs

ddurieux:


alignak/objects/schedulingitem.py
```diff
@@ -1763,6 +1763,11 @@ def consume_result(self, chk, notif_period, hosts,  # pylint: disable=R0915,R091
             if self.state_type == 'SOFT':
                 if not chk.is_dependent():
                     self.add_attempt()
+                # Cases where go:
+                #  * warning soft => critical hard
```
Seb-Solon: If we are in critical hard, then we are not in this case (see above, state_type == SOFT)

ddurieux: It's normal, it's 5 lines after we put SOFT in HARD state because max attempt

alignak/external_command.py
```diff
@@ -1122,9 +1122,9 @@ def acknowledge_host_problem_expire(self, host, sticky, notify,
         TODO: add a better ACK management
         """
         notif_period = self.daemon.timeperiods[host.notification_period]
-        self.send_an_element(host.acknowledge_problem(notif_period, None, sticky, notify,
-                                                      persistent, author,
-                                                      comment, end_time=end_time))
+        self.sched.add(host.acknowledge_problem(notif_period, self.hosts, self.services, sticky,
```
mohierf: Please use `send_an_element` function ...

test/test_acknowledge.py
```diff
@@ -0,0 +1,887 @@
...
+        self.scheduler_loop(1, [[host, 2, 'DOWN']])
+        time.sleep(0.1)
+        self.assertEqual("DOWN", host.state)
+        self.assertEqual("SOFT", host.state_type)
+
+        now = time.time()
+        cmd = "[{0}] ACKNOWLEDGE_HOST_PROBLEM;{1};{2};{3};{4};{5};{6}\n".\
+            format(now, host.host_name, 2, 0, 1, 'darth vader', 'normal process')
+        self.schedulers['scheduler-master'].sched.run_external_command(cmd)
+
```
mohierf: Why not testing immediately that: self.assertTrue(host.problem_has_been_acknowledged) ?

test/test_acknowledge.py
```diff
@@ -0,0 +1,887 @@
...
+        self.scheduler_loop(1, [[host, 0, 'UP'], [host_router, 0, 'UP'], [svc, 0, 'OK']])
+        time.sleep(0.1)
+        self.assertFalse(host.problem_has_been_acknowledged)
+        self.assert_actions_count(0)
+
+        self.scheduler_loop(1, [[host_router, 2, 'DOWN']], False)
+        time.sleep(0.1)
+        self.assertEqual("DOWN", host_router.state)
+        self.assertEqual("SOFT", host_router.state_type)
+        self.assertEqual("UP", host.state)
```
mohierf: host should be unreachable, no ?

ddurieux: No because I have disabled the dependencies

test/test_acknowledge.py
ddurieux: I do it few lines after

PR #393, opened 2016-10-03
Add test for command multilines (like command with html code)

ddurieux:


alignak/objects/config.py
```diff
@@ -1103,13 +1103,14 @@ def read_config_buf(self, buf):  # pylint: disable=R0912
 
             line_nb += 1
             # Remove comments
-            line = split_semicolon(line)[0].strip()
```
Seb-Solon: So we consider semicolon on continuation line as legit char ? This affect all file parsing here, carful

alignak/objects/config.py
```diff
@@ -1103,13 +1103,14 @@ def read_config_buf(self, buf):  # pylint: disable=R0912
...
+            if not continuation_line:
+                line = split_semicolon(line)[0].strip()
 
             # A backslash means, there is more to come
             if re.search(r"\\\s*$", line) is not None:
                 continuation_line = True
                 line = re.sub(r"\\\s*$", "", line)
-                line = re.sub(r"^\s+", " ", line)
+                # line = re.sub(r"^\s+", " ", line)
```
Seb-Solon: Why comment this? We are replacing line by only one space. This affect all file parsing here, carful

alignak/objects/config.py
mohierf: I agree with @Seb-Solon ... we should not remove this line because it is used to remove comments starting with a semi-colon on the same line.

ddurieux:
Ok I close

PR #396, opened 2016-10-03
Add property hostgroup_members in Hostgroup class (same for servicegr…

### LINKED ISSUE #110: Add property hostgroup_members in Hostgroup class

ddurieux:
…oup and contactgroup). closes #110


alignak/objects/hostgroup.py
```diff
@@ -112,18 +114,13 @@ def get_hosts(self):
...
         :rtype: list
         """
-        # TODO: consistency : a Hostgroup instance should always
-        # have its hostgroup_members attribute defined, even if the empty list
         if hasattr(self, 'hostgroup_members'):
-            # consistency: any Hostgroup instance's hostgroup_members attribute
-            # should already be decoded/parsed:
-            # this should already be in its list form.
-            return [m.strip() for m in self.hostgroup_members.split(',')]
+            return self.hostgroup_members
```
Seb-Solon: Do you ensure members are strip at parsing?

alignak/objects/contactgroup.py
```diff
@@ -73,6 +73,8 @@ class Contactgroup(Itemgroup):
         'uuid':                 StringProp(default='', fill_brok=['full_status']),
         'contactgroup_name':    StringProp(fill_brok=['full_status']),
         'alias':                StringProp(fill_brok=['full_status']),
+        'contactgroup_members': ListProp(default=[], fill_brok=['full_status'],
```
Seb-Solon: we should avoid this. Unless you are 100% sure this value is replaced or not use to append something. 
For example add_string_member function dos not ensure that (hasattr will return true) but add_string_unknown_member have this (bool([]) is false) 

See #403

alignak/objects/hostgroup.py
```diff
@@ -71,14 +71,16 @@ class Hostgroup(Itemgroup):
...
-        'alias':          StringProp(fill_brok=['full_status']),
-        'notes':          StringProp(default='', fill_brok=['full_status']),
-        'notes_url':      StringProp(default='', fill_brok=['full_status']),
-        'action_url':     StringProp(default='', fill_brok=['full_status']),
-        'realm':          StringProp(default='', fill_brok=['full_status'],
-                                     conf_send_preparation=get_obj_name),
+        'uuid':                 StringProp(default='', fill_brok=['full_status']),
+        'hostgroup_name':       StringProp(fill_brok=['full_status']),
+        'alias':                StringProp(fill_brok=['full_status']),
+        'hostgroup_members':    ListProp(default=[], fill_brok=['full_status'],
```
Seb-Solon: see previous comment

mohierf:
We should merge this PR because it breaks some configuration with hostgroups  in hosts or services dependencies ...

alignak/objects/contactgroup.py
mohierf: I propose to leave as is ... this code is running ok actually.

mohierf: There is a syntax error on the line you added ... missing parenthesis to close the ListProp!

Seb-Solon:
"this code is running ok actually" is not a really legit excuse to merge a PR. It looks like a "it's works on my machine" answer.  Why merging if we know that it can cause issue?

mohierf:
I do agree ... but some tests are still existing and some others are yet to come ... ddurieux needs to make some more PR for the tests that were in the former develop+enhanceunitests branch.

I understand your comment on the problems that may appear because of [] initializing but we do not match this case currently. And the bug in the groups members was a real problem in our test configuration ... and I am fed-up with making some patches to tests more important features than this one...

PR #425, opened 2016-10-07
Clean alignak_test

ddurieux:


test/alignak_test.py
```diff
@@ -272,51 +272,10 @@ def fake_check(self, ref, exit_status, output="OK"):
...
-                chk.set_type_active()
-                chk.output = output
-                chk.exit_status = exit_status
-                self.schedulers['scheduler-master'].sched.waiting_results.put(chk)
-            for i in self.schedulers['scheduler-master'].sched.recurrent_works:
-                (name, fun, nb_ticks) = self.schedulers['scheduler-master'].sched.recurrent_works[i]
-                if nb_ticks == 1:
-                    fun()
-
     def scheduler_loop_new(self, count, items):
```
mohierf: Why not renaming it as scheduler_loop? Why leaving the _new ...

ddurieux: You're right, I fix that

Issue #426, opened 2016-10-07
Arbiter daemon configuration


mohierf:
Currently the arbiter daemon configuration is located in the `alignak.cfg` file ... thus:
- mixing the daemon process configuration and the monitoring configuration
- making it quite complicated to configure the arbiter log
- creating a specific case for this daemon in the log configuration process

Because the whole monitoring configuration needs to be loaded and parsed before being able to configure the logger, this obliges to store intermediate logs ... and makes the log processing very specific and overloaded. Each log needs to make a test before emitting the log to the logger.

Linked with #424

Issue #424, opened 2016-10-07
Arbiter own configuration file


mohierf:
Having an `arbiterd.ini` file like for other daemons will allow to improve the log management. This will avoid to store the logs emitted before the alignak configuration parsing ...

mohierf:
The arbiter daemon can have its own configuration file `arbiterd.ini` which will be installed with the defautt installation process.

The arbiter command line parameters parsing must be modified as is: `alignak-arbiter -V -c /usr/local/etc/alignak/daemons/arbited.ini -c /usr/local/etc/alignak/alignak.cfg`. The first configuration file will be considered as the daemon configuration file (eg. arbiterd.ini) and the following files will be considered as the monitoring configuration files (eg. alignak.cfg).

If only one configuration file is provided, it is considered as the monitoring configuration file (eg. alignak.cfg). This to achieve compatibility with existing installations and configurations ...

The default values, if no daemon configuration file exists are modified to match with the dafault arbiter configuration (port number, log file name, ...)

mohierf:
With more brainstorming, it would be more interesting to have some default parameters in the Arbiter daemon and to create a new command line parameter (eg. `-m --monitored`) to specify the monitoring configuration files

What about this @ddurieux @Seb-Solon ?

Issue #414, opened 2016-10-05
Alignak arbiter verify mode logs


mohierf:
The Alignak arbiter verify mode do not log less than WARNING. All the INFO log are no more displayed when running `alignak-arbiter -V -c /usr/local/etc/alignak/alignak.cfg`

mohierf:
We should set the global Alignak logger default to INFO in the arbiter damon to be sure that the configuration loading is logged.

mohierf:
Moreover, the configuration check logs only use the Console handler ...

Issue #413, opened 2016-10-05
Default log level to INFO


mohierf:
The default log level for all the daemons should be set to INFO instead of WARNING

PR #427, opened 2016-10-08
Develop+logs

mohierf:
Refactor log module:
- remove Log class (and BrokHandler, and logs stacking)
- replace with a setup function that allows to define a Console and a RotatingFile handler
- allow to change log level and date format
- replace naglog_result function with make_monitoring_log function that returns a Brok
- allow to configure logger in the daemon configuration file

**Impacts:**
- base Daemon class and daemons logger configuration modified
- Arbiter now has its own arbiterd.ini file
- naglog_result is replaced with a function creating a Brok 

This pull request completes #386, replaces #411 and is linked to:
- #424 and #426: the Arbiter has its own configuration file as other daemons
- #409, #410, #413 and #414 for the logs refactoring process

### DIFF
--- a/alignak/bin/alignak_arbiter.py
+++ b/alignak/bin/alignak_arbiter.py
@@ -65,8 +65,8 @@ def main():
     """
     args = parse_daemon_args(True)
 
-    if not args.config_files:
-        print "Requires at least one config file (option -c/--config"
+    if not args.monitoring_files:
+        print "Requires at least one monitoring configuration file (option -a/--arbiter)"
         sys.exit(2)
 
     # Protect for windows multiprocessing that will RELAUNCH all
@@ -76,7 +76,6 @@ def main():
         if not daemon.need_config_reload:
             break
         daemon = None
-
 
 if __name__ == '__main__':
     main()
--- a/alignak/brok.py
+++ b/alignak/brok.py
@@ -58,6 +58,24 @@ from alignak.misc.serialization import serialize, unserialize, AlignakClassLooku
 class Brok(object):
     """A Brok is a piece of information exported by Alignak to the Broker.
     Broker can do whatever he wants with it.
+
+    Broks types:
+    - log
+    - monitoring_log
+
+    - notification_raise
+    - downtime_raise
+    - initial_host_status, initial_service_status, initial_contact_status
+    - initial_broks_done
+
+    - update_host_status, update_service_status, initial_contact_status
+    - host_check_result, service_check_result
+    - host_next_schedule, service_next_scheduler
+    - host_snapshot, service_snapshot
+    - unknown_host_check_result, unknown_service_check_result
+
+    - program_status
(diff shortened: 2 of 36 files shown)


alignak/bin/alignak_arbiter.py
```diff
@@ -65,17 +65,13 @@ def main():
...
 
     # Protect for windows multiprocessing that will RELAUNCH all
-    while True:
-        daemon = Arbiter(debug=args.debug_file is not None, **args.__dict__)
-        daemon.main()
-        if not daemon.need_config_reload:
-            break
-        daemon = None
+    daemon = Arbiter(debug=args.debug_file is not None, **args.__dict__)
+    daemon.main()
```
ddurieux: With remove the code, you break the reload process

mohierf: There was a mix-up with another concurring PR ... sorry

mohierf: It would be better to have a `return` rather than a ` break`, no?

mohierf: We are not in a while ... but in a main function ;)

alignak/daemon.py
```diff
@@ -126,6 +126,7 @@ def get_all_groups():
         """
         return []
 
+from alignak.log import logger, setup_logger, get_logger_fds
```
Seb-Solon: you should not import logger here but recreate it like on the other modules.

alignak/brok.py
```diff
@@ -58,6 +58,24 @@
...
+    - initial_broks_done
+
+    - update_host_status, update_service_status, initial_contact_status
+    - host_check_result, service_check_result
+    - host_next_schedule, service_next_scheduler
+    - host_snapshot, service_snapshot
+    - unknown_host_check_result, unknown_service_check_result
+
+    - program_status
+    - clean_all_my_instance_id
```
mohierf: Ok

alignak/daemon.py
mohierf: Ok. Importation is of no effect because we instantiate a logger some few lines later...

alignak/daemon.py
```diff
@@ -146,12 +147,11 @@ def get_all_groups():
...
 
 class InvalidPidFile(Exception):
-    """Exception raise when a pid file is invalid"""
+    """Exception raised when a pid file is invalid"""
     pass
 
 
-DEFAULT_WORK_DIR = '/var/run/alignak/'
-DEFAULT_LIB_DIR = '/var/lib/alignak/'
+DEFAULT_WORK_DIR = './'
```
mohierf: This is not even useful because it is only used to initialize a property ... and when the property is used we os.chdir the current directory

mohierf:
@Seb-Solon only consider the last commit (ignore the first one which is a mistake due to my messay configuration this morning ...)

Issue #451, opened 2016-10-12
Daemon / modules loading logs and improvements


mohierf:
When the daemons are starting, many log are sent but some are not really useful for the end-user to understand how the configuration is dispatched, some logs are missing to help user understand if his configuration is dispatched and operationnal: did all the daemons received their part and which daemon is connected to any other daemon.

This is because the logs are not sent equally by all the daemons

mohierf:
Will clean #447

mohierf:
Closed with #453

PR #453, opened 2016-10-12
Daemons / modules loading:

### LINKED ISSUE #251: Stop a daemon with modules
When a daemon using modules is stopped, the external modules are killed abruptly ... because there is no stop negotiation between daemon and its modules. The daemon can not send any signal to kill its modules and then it kills the modules after few seconds ... this does not allow an external module to end properly!

This behaviour should be changed because some external modules may:
- be aware that someone required a shutdown
- have something to finish before ending

EDIT:
Found in the logs:
```
[1447852514] WARNING: [broker-master] u'alignakbackendbrok' is still alive normal kill, I help it to die
```

mohierf:
- #451: logs cleaning
  - module loading improvement
  - few code cleaning (attributes names)

Closes #251


ddurieux:
Seems good idea to rename function _get_name_ in _get_alias_

alignak/daemons/brokerdaemon.py
```diff
@@ -509,7 +510,10 @@ def setup_new_conf(self):  # pylint: disable=R0915,R0912
                 self.schedulers[sched_id]['timeout'] = sched['timeout']
                 self.schedulers[sched_id]['data_timeout'] = sched['data_timeout']
 
-            logger.info("We have our schedulers: %s ", self.schedulers)
+            logger.debug("We have our schedulers: %s", self.schedulers)
+            logger.info("We have our schedulers:")
+            for sat_id in self.schedulers:
+                logger.info(" - %s ", self.schedulers[sat_id]['name'])
```
ddurieux: it's more simple to do:
```
for sched in self.schedulers.values():
    logger.info(" - %s ", sched['name'])
```

alignak/basemodule.py
```diff
@@ -246,12 +245,14 @@ def stop_process(self):
         :return: None
         """
         if self.process:
-            logger.info("I'm stopping module %r (pid=%s)",
+            logger.info("I'm stopping module %r (pid=%d)",
                         self.get_name(), self.process.pid)
             self.process.terminate()
+            # Todo: get sure that 1 second is enough ...
```
mohierf: You are right. I remove the timeout ...

alignak/daemons/brokerdaemon.py
mohierf: +1

alignak/basemodule.py
mohierf: I'd rather let the timeout, and even increase it to 10 seconds, ... because if the process did not terminated after the timeout we force kill it ... then it sees better to let the child process some time to terminate properly before killing it.

alignak/basemodule.py
```diff
@@ -246,12 +245,14 @@ def stop_process(self):
         :return: None
         """
         if self.process:
-            logger.info("I'm stopping module %r (pid=%s)",
+            logger.info("I'm stopping module %r (pid=%d)",
                         self.get_name(), self.process.pid)
             self.process.terminate()
-            self.process.join(timeout=1)
+            # Todo: get sure that 1 second is enough ...
+            # some external modules may need more time to stop, no?
```
ddurieux: Need update the comment

mohierf: :) updated

PR #460, opened 2016-10-13
Arbiter warnings/errors raised and reported when reading files

### LINKED ISSUE #459: Invalid configuration detected but no information provided
Sometimes, an invalid configuration is detected by the Arbiter when reading the configuration files, but no information is provided on the console nor in the log ...

mohierf:
Closes #459


alignak/objects/config.py
```diff
@@ -827,11 +827,9 @@ def clean_params(self, params):
         for elt in params:
             elts = elt.split('=', 1)
             if len(elts) == 1:  # error, there is no = !
-                self.conf_is_correct = False
```
ddurieux: Are you sure we can remove self.conf_is_correct = False ?

mohierf: Yes of course ... it is set in the `self.add_error` function :wink:

PR #468, opened 2016-10-16
Add new brok: new_conf. closes #466

### LINKED ISSUE #466: Add new brok type: new_conf
This _brok_ will be created by the scheduler when it receive and manage the new conf. 
The goal is to give the information to the external brokers modules that the scheduler has a new configuration so update / new objects

ddurieux:


alignak/daemons/schedulerdaemon.py
```diff
@@ -335,6 +336,10 @@ def setup_new_conf(self):
             # and set ourselves in it
             self.schedulers = {self.conf.uuid: self.sched}  # pylint: disable=E1101
 
+            # Create brok new conf
+            brok = Brok({'type': 'new_conf', 'data': {}})
```
mohierf: Why not adding the configuration id ?

ddurieux: Is it useful the flavour id ?

PR #469, opened 2016-10-17
Alignak modules management

mohierf:
Alignak module loading:
  - identify Nagios parameters requiring external modules and bail out if modules are not existing
  - add the module_types property for the modules (retention, logs, livestate, ...)
  - check the module loading and properties to validate the Alignak configuration
  - clean the module loading
  - clean the module logging

Modules unit tests:
  - extend AlignakTest properties
  - modules testing and loading unit tests

Update virtualenv tests:
 - update test requirements
 - update files lists

**Note:** I do not understand why I had to modify the virtual env tests ... but it was necessary to make the Travis build become Ok! Thus, I do not une derstand why thoses tests are ok in the current develop branch !!!


alignak/basemodule.py
```diff
@@ -99,6 +100,7 @@ def __init__(self, mod_conf):
         """
         self.myconf = mod_conf
         self.alias = mod_conf.get_name()
+        # Todo: disabled feature
```
ddurieux: disabled what feature? not understand

mohierf:
I foudn an extra logger.error in the source code that I will remove ... here: https://github.com/Alignak-monitoring/alignak/blob/example-module/alignak/modulesmanager.py#L247

mohierf:
I also found that it would be better to change the default `init` function for it returns True as a defaut behaviour ... I will update.

alignak/modulesmanager.py
```diff
@@ -143,58 +121,85 @@ def load(self, modules):
...
-                logger.warning("Module %s (%s) can't be loaded because attributes errors",
-                               module.python_name, module.module_alias)
-                logger.exception("Exception: %s", exp)
+                self.configuration_errors.append(
+                    "Module %s (%s) can't be loaded, Python importation error: %s" %
+                    (module.python_name, module.module_alias, str(exp))
+                )
+            except AttributeError:
+                self.configuration_errors.append(
+                    "Module %s (%s) can't be loaded, module configuration" %
```
ddurieux: Perhaps add more info in log like _properties and/or get_instance_ missing to help the user / developer

alignak/objects/config.py
```diff
@@ -1866,152 +1863,124 @@ def create_business_rules_dependencies(self):
...
+                      ('use_syslog', self.use_syslog)
+                logger.error(msg)
+                self.configuration_errors.append(msg)
+            else:
+                msg = "Your configuration parameters '%s = %s' are deprecated " \
+                      "and will be ignored. Please configure your external 'logs' module " \
+                      "as expected." % \
+                      ('use_syslog', self.use_syslog)
+                logger.warning(msg)
+                self.configuration_warnings.append(msg)
```
ddurieux: these 13 line before, you can merge with little before and test _log_file_ because same code for _log_file_ and _use_syslog_

alignak/modulesmanager.py
mohierf: this is already logged before ...

alignak/objects/config.py
mohierf: Intentional code duplicate because every parameter has its own message ...

PR #480, opened 2016-10-19
Fix #479: missing contact methods

### LINKED ISSUE #479: Missing functions in Contact class for macros
The method `get_groupname` and `get_groupnames` used for the contact macros CONTACTGROUPNAME and CONTACTGROUPNAMES

mohierf:
I simply fixed in to Alignak what I proposed as a Shinken PR a while ago...

### DIFF
--- a/alignak/objects/contact.py
+++ b/alignak/objects/contact.py
@@ -218,6 +218,22 @@ class Contact(Item):
         except AttributeError:
             return 'UnnamedContact'
 
+    def get_groupname(self):
+        """
+        Get the first group name whose contact belongs to
+        :return: group name
+        :rtype: str
+        """
+        return self.contactgroups[0]
+
+    def get_groupnames(self):
+        """
+        Get all the groups name whose contact belongs to
+        :return: comma separated list of the groups names
+        :rtype: str
+        """
+        return ', '.join(self.contactgroups)
+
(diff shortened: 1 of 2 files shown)


alignak/objects/contact.py
```diff
@@ -182,6 +182,33 @@ def get_name(self):
...
+        Get the first group name whose contact belongs to
+        :return: group name
+        :rtype: str
+        """
+        groupname = ''
+        for contactgroup in self.contactgroups:
+            groupname = contactgroup
+            break
+
+        return groupname
```
Seb-Solon:
```
if self.contactgroups:
    return self.contactgroups[0]
else
    return ''
```

for loop is unnecessary

alignak/objects/contact.py
```diff
@@ -182,6 +182,33 @@ def get_name(self):
...
+        :return: comma separated list of the groups names
+        :rtype: str
+        """
+        groupnames = ''
+        for contactgroup in self.contactgroups:
+            if groupnames == '':
+                groupnames = contactgroup
+            else:
+                groupnames = "%s, %s" % (groupnames, contactgroup)
+        return groupnames
```
Seb-Solon:
```
return ', '.join(self.contactgroups)
```

for loop is unnecessary

mohierf: You are right :wink:

PR #499, opened 2016-10-24
WiP: Send a notification if a parent is down 

mohierf:
we require 'u' notifications ...


hvad:
Hi,
I don't think this is a good idea and it's not a normal usage of tools nagios like. And can send you a email storm when network outage. 
Bye

mohierf:
Abandoned

Issue #514, opened 2016-10-27
Alignak test suite is broken


mohierf:
Some tests in the unit tests suite are not executed by the Travis process.

Indeed, when a test script has an execute attribute (755), the nosetests executable ignore this file ! Thus some tests were not executedx by the Travis CI...

mohierf:
Fixed with #515

PR #515, opened 2016-10-27
Fix files rights and, thus, broken tests

mohierf:
Because of #514 the unit test suite is broken. This PR:
 - fixes Alignak python files access rights
 - fixes the broken tests
 - improves the code coverage thanks to running code coverage inside the running daemons

**Edit**: I intentionnaly left separated commits ... this will not be too annoying for the review :wink:


test/test_config.py
```diff
@@ -472,35 +461,38 @@ def test_bad_realm_conf(self):
...
         )
         self.assert_any_cfg_log_match(
-            "Error: the realm configuration of yours hosts is not good because there is more "
-            r"than one realm in one pack \(host relations\):"
+            "realms configuration is incorrect!"
         )
         self.assert_any_cfg_log_match(
-            "the host test_host_realm2 is in the realm Realm2"
+            re.escape(
+                "Error: Hosts exist in the realm <Realm \"name\"=u'Realm2' /> "
```
ddurieux: It could be better to have the realm name instead the object Realm in the log

alignak/objects/host.py
```diff
@@ -291,7 +291,7 @@ def fill_predictive_missing_parameters(self):
             self.alias = self.host_name
         if self.initial_state == 'd':
             self.state = 'DOWN'
-        elif self.initial_state == 'x':
+        elif self.initial_state in ['u', 'x']:
```
mohierf: Ok

test/test_config.py
mohierf: It is one of your modification that reintroduced the object rather than the name :wink: 
I will replace with the realm name...

test/test_modules.py
```diff
@@ -209,9 +207,10 @@ def test_module_on_module(self):
         modules = [m.module_alias for m in self.schedulers['scheduler-master'].modules]
         self.assertListEqual(modules, ['Example'])
 
-    @unittest.skip("To make a test with Travis")
+    # @unittest.skip("To make a test with Travis")
```
ddurieux: So you can delete this line ;)

PR #516, opened 2016-10-27
Add ssl test and fix cherrypy start with SSL. closes #502

### LINKED ISSUE #502: Add a unit tests for SSL
Add a unit test like full_tst.py with an SSL configuration

ddurieux:


alignak/daemon.py
```diff
@@ -182,7 +182,9 @@ class Daemon(object):  # pragma: no cover, not for unit tests...
         'server_key':
             StringProp(default='etc/certs/server.key'),
         'ca_cert':
-            StringProp(default='etc/certs/ca.pem'),
+            StringProp(default=''),
+        'server_dh':
```
Seb-Solon: What is that for?

alignak/http/daemon.py
```diff
@@ -46,6 +46,51 @@
...
+            cont = SSL.Context(SSL.SSLv23_METHOD)
+
+            # override:
+            ciphers = (
+                'ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+HIGH:'
+                'DH+HIGH:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+HIGH:RSA+3DES:!aNULL:'
+                '!eNULL:!MD5:!DSS:!RC4:!SSLv2'
+            )
+            cont.set_options(SSL.OP_NO_COMPRESSION | SSL.OP_SINGLE_DH_USE | SSL.OP_NO_SSLv2 |
+                             SSL.OP_NO_SSLv3)
```
Seb-Solon: Previously it was : SSL.OP_NO_SSLv2 | SSL.OP_NO_SSLv3 | SSL.OP_NO_TLSv1 | SSL.OP_NO_TLSv1_1 |        SSL.OP_NO_COMPRESSION. Why allowing TLSv1 and TLSv1.1 ? What is SINGLE DH ?

alignak/http/daemon.py
```diff
@@ -46,6 +46,51 @@
...
+
+        def get_context(self):
+            """Return an SSL.Context from self attributes."""
+            cont = SSL.Context(SSL.SSLv23_METHOD)
+
+            # override:
+            ciphers = (
+                'ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+HIGH:'
+                'DH+HIGH:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+HIGH:RSA+3DES:!aNULL:'
+                '!eNULL:!MD5:!DSS:!RC4:!SSLv2'
```
Seb-Solon: What is the differnce with 'DEFAULT:!DSS:!PSK:!SRP:!3DES:!RC4:!DES:!IDEA:!RC2:!NULL' ?
Here you are manually listing what you want and then still remove what you don't want.

alignak/daemon.py
ddurieux: see  https://wiki.openssl.org/index.php/Diffie_Hellman
It's used to better secure the SSL transaction

mohierf:
@ddurieux: I propose a daemon configuration file modification for SSL:
 - add an `etcdir=/usr/local/etc/alignak` variable (like `workdir` and `logdir`)
 - replace */usr/local/etc/alignak* with *%(etcdir)* in the SSL properties `ca_cert`, `server_cert` and `server_key`

Change all daemons configuration file with this please

mohierf:
@ddurieux : another suggestion: include default SSL configuration files in the Alignak default configuration according to the default values of the daemon configuration files. Those files should be located in *etc/certs* directory...

ddurieux:
No, people MUST generate their certificate and we don't offer them it because it's a security problem, I prefer add doc with commands to generate it

test/test_ssl.py
```diff
@@ -0,0 +1,129 @@
...
+        shutil.copy('cfg/ssl/certificate_test.key', '/tmp/')
+        shutil.copy('cfg/ssl/dhparams.pem', '/tmp/')
+        self.procs = {}
+        self.ssl_installed = True
+        try:
+            from OpenSSL import SSL
+        except ImportError:
+            self.ssl_installed = False
+            print "Install pyopenssl"
+            subprocess.call("pip install pyopenssl", shell=True)
```
Seb-Solon: Add this to the test-requirement Instead of dynamically calling pip.

ddurieux: The goal is to have other tests works without pyopenssl not installed and run this test with pyopenssl. Like this we test with and without (it's a good idea because yesterday without pyopenssl alignak has broken, but I have fixed it)

Issue #517, opened 2016-10-28
Daemons do not parse the ini files correctly


mohierf:
Some ConfigParser variables used for interpolation are present in the configuration files (eg. _%(workdir)s_, ...) but they are not interpolated on configuration load ...

As an exemple: set this in _schedulerd.ini_:

```
logdir=/usr/local/var/log/alignak
local_log=%(logdir)s/schedulerd.log
```

We do not have problems currently because the installation script replaces the `local_log` variable... but it should not! So there is also a problem with the setup script... bug AND bug = NO bug :wink:

PR #520, opened 2016-10-28
Daemons configuration (#517):

mohierf:
- update default configuration files
- catch daemon PortNotFree exception
- add a test for daemon starting


.travis.yml
```diff
@@ -33,7 +33,7 @@ after_success:
   # to get coverage data with relative paths and not absolute we have to
   # execute coveralls from the base directory of the project,
   # so we need to move the .coverage file here :
-  - if [[ $TEST_SUITE == 'unit' ]]; then mv test/.coverage . && coveralls --rcfile=test/.coveragerc; fi
+  - if [[ $TEST_SUITE == 'unit' ]]; then mv test/.coverage . && coveralls debug && coveralls -v --rcfile=test/.coveragerc; fi
```
ddurieux: We can't remove debug ?

alignak/http/daemon.py
```diff
@@ -62,17 +62,37 @@ class HTTPDaemon(object):
...
     def __init__(self, host, port, http_interface, use_ssl, ca_cert,
                  ssl_key, ssl_cert, daemon_thread_pool_size):
-        self.port = port
-        self.host = host
-        self.srv = None
+        """
+        Initialize HTTP daemon
+
+        :param host: host address
+        :param port: listening port
```
ddurieux: add:
```:type port: int```

.travis.yml
mohierf: Yes we can ... but it remains an interesting information, no ? This is why I left for the moment ...

alignak/http/daemon.py
mohierf: I added in my next PR ;) It may wait, no ?

PR #530, opened 2016-10-31
Satellites need to know all the Alignak objects...

mohierf:
... else it is not possible to unserialize poller/reactionne modules.

### DIFF
--- a/alignak/satellite.py
+++ b/alignak/satellite.py
@@ -72,10 +72,6 @@ import time
 import traceback
 import threading
 
-# pylint: disable=wildcard-import,unused-wildcard-import
-# This import, despite not used, is necessary to include all Alignak objects modules
-from alignak.objects import *
-
 from alignak.http.client import HTTPClient, HTTPEXCEPTIONS
 from alignak.http.generic_interface import GenericInterface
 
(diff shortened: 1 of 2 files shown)


Seb-Solon:
Can you provide an example ?

We have added such thing in the init file of bin lib (https://github.com/Alignak-monitoring/alignak/blob/develop/alignak/bin/__init__.py). In what case we use satellite without a daemon?

mohierf:
Example is very easy: simply add a module to a poller and you will have an AlignakClassLookupException exception when the poller tries to unserialize its modules ...

Seb-Solon:
Yes, but what the stack trace is? Do you think, it is only a matter of adding the missing object into the __init__ file I mentioned?

mohierf:
You got it :+1:

alignak/satellite.py
```diff
@@ -995,16 +999,18 @@ def setup_new_conf(self):  # pylint: disable=R0915,R0912
...
-            mods = unserialize(g_conf['modules'], True)
-            self.new_modules_conf = []
-            for module in mods:
-                # If we already got it, bypass
-                if module.python_name not in self.q_by_mod:
-                    logger.debug("Add module object %s", str(module))
-                    self.new_modules_conf.append(module)
-                    logger.info("[%s] Got module: %s ", self.name, module.python_name)
-                    self.q_by_mod[module.python_name] = {}
+            if not self.have_modules:
```
Seb-Solon: Why is this if needed?

mohierf: It is not necessary. I added this when testing but I will remove ...

mohierf:
@Seb-Solon can you merge this PR please, else the nrpe-booster can not be used with Alignak ... :/P

Seb-Solon:
NB : This PR remove some detail in debug logs to be discussed later #535

PR #533, opened 2016-11-03
[WiP] Fix #532: add realm name in the satellites

### LINKED ISSUE #532: Arbiter HTTP interface provides uuid for realm property
The Arbiter HTTP interface do not provide the realm name for the realms but the uuid:
```
reactionner: {
reactionner-master: {
passive: false,
realm: "d37a81a54f784b8e9d6c0812a4cc98b5",
polling_interval: 1,
alive: true,
manage_arbiters: false,
manage_sub_realms: false,
is_sent: true,
spare: false,
check_interval: 60,
address: "127.0.0.1",
reachable: true,
max_check_attempts: 3,
last_check: 1478093870.726019,
port: 7769
}
},
```

mohierf:
This to allow the arbiter API to provide a realm uuid **and** a realm name

### DIFF
--- a/alignak/http/arbiter_interface.py
+++ b/alignak/http/arbiter_interface.py
@@ -163,8 +163,8 @@ class ArbiterInterface(GenericInterface):
                         if not hasattr(daemon, prop):
                             continue
                         val = getattr(daemon, prop)
-                        if prop == "realm" and hasattr(val, "realm_name"):
-                            env[prop] = val.realm_name
+                        if prop == "realm":
+                            continue
                         # give a try to a json able object
                         try:
                             json.dumps(val)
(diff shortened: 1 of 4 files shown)


Seb-Solon:
Summary of the discussion we had on IRC: uuid should end up in the backend to retrieve the full object if necessary. The current patch is good enough because there as no requirements / usage defined on this http endpoint. If we need more information later we may fetch it internally (self.app is the arbiter daemon)

mohierf:
Not sure that the best solution is to push the object `uuid` to the backend ... because `uuid` is an internal Alignak one and the backend has its own identifier for the stored objects. To be discussed ...

Seb-Solon:
The backend should have this extra information stored even if it is not called uuid. Alignak object are linked with those id interenally you may need this uuid in other cases.

From what I remember object are identified by *_name attribute (the only was before), there are some work to support the alignak uuid now.

Issue #541, opened 2016-11-05
Reactionner module loading exception


mohierf:
Same problem as the poller one: #529 An exception is raised in the reactionner when it tries to unserialize the modules it got in its configuration ...

mohierf:
Moreover it seems that the problem potentially fixed with #529 still occur in some conditions ... I got it again in a test module !

Seb-Solon:
Paste exception please so that we can follow

mohierf:
It is always the same and only one when unserializing: `AlignakClassLookupException` :wink:

Seb-Solon:
The interesting part is which class was not found to be able to add it into the list

mohierf:
Always the same also in this case as explained in the first issue comment: *when it tries to unserialize the modules it got in its configuration ...*

mohierf:
Fixed with #543

Seb-Solon:
The stack trace + exception will always remove any doubt.  "when it tries to unserialize the modules" could raise another exception (let's imagine the Module object has other objects embedded, on this one is faulty?)

Anyway the bug is fixed, I was just trying to fully understand the issue

PR #542, opened 2016-11-05
[WiP] satellites and modules fixes and tests

### LINKED ISSUE #529: Alignak poller do not load its modules
When defining a poller module, an exception is raised when the poller receives its configuration. The poller cannot unserialize the received modules ...

### LINKED ISSUE #539: Alignak satellites workers queues named as module python_name
The Alignak satellites workers queues are named with the module `python_name` attribute. 

This is a problem with some modules like NRPE booster or SNMP booster because the commands that are attached to a module must be tagged with something like `module_type alignak_module_nrpe_booster` or  `module_type alignak_module_snmp_booster` ...

**Note**: Formerly, satellites used the module `type` property (eg. nrpe_poller, snmp_poller) and we must restore this behavior

mohierf:
Fix #539: satellites queues named as module python_name
Fix #529 and #541: module loading exception in poller and reactionner
Add a daemons with modules launch test
Fix test files rights to 664


mohierf:
Replaced with most recent #543 because of a bad branch name :disappointed:

PR #543, opened 2016-11-05
Modules and satellites

### LINKED ISSUE #529: Alignak poller do not load its modules

### LINKED ISSUE #539: Alignak satellites workers queues named as module python_name

mohierf:
Fix #539: satellites queues named as module python_name
Fix #529 and #541: module loading exception in poller and reactionner
Add a daemons with modules launch test
Fix test files rights to 664

-----

Tested also on demo server


alignak/satellite.py
```diff
@@ -1000,11 +1002,12 @@ def setup_new_conf(self):  # pylint: disable=R0915,R0912
             self.new_modules_conf = []
             for module in mods:
                 # If we already got it, bypass
-                if module.python_name not in self.q_by_mod:
+                if module.get_name() not in self.q_by_mod:
+                    logger.warning("Add module object: %s", module)
```
ddurieux: Prefer use info instead warning

mohierf: You are right ... INFO will be better. I fix this.

test/test_daemon_start.py
```diff
@@ -195,6 +198,26 @@ def test_config_and_start_and_stop(self):
         self.stop_daemon(d)
         self.assertFalse(os.path.exists(d.pidfile))
 
+        # Start as a daemon
```
Seb-Solon: comment does not match the real call

Seb-Solon: You are a doing two time the same thing actually. The following line are copy paste from the previous ones.

test/test_launch_daemons.py
```diff
@@ -59,6 +59,119 @@ def setUp(self):
...
+                os.remove('/tmp/%sd.pid' % daemon)
+                print("- removed /tmp/%sd.pid" % daemon)
+            if os.path.exists('/tmp/%sd.log' % daemon):
+                os.remove('/tmp/%sd.log' % daemon)
+                print("- removed /tmp/%sd.log" % daemon)
+
+        print("Launching arbiter with bad configuration file...")
+        args = ["../alignak/bin/alignak_arbiter.py",
+                "-V",
+                "-a", "cfg/run_test_launch_daemons/alignak.cfg"]
```
Seb-Solon: ini is not needed with -V ?

alignak/satellite.py
```diff
@@ -83,6 +83,7 @@
 from alignak.daemon import Daemon
 from alignak.stats import statsmgr
 from alignak.check import Check  # pylint: disable=W0611
+from alignak.objects.module import Module  # pylint: disable=W0611
```
Seb-Solon: This import is already imported here : https://github.com/Alignak-monitoring/alignak/blob/develop/alignak/bin/__init__.py. I don't think this was necessary.

test/test_daemon_start.py
```diff
@@ -123,18 +124,19 @@ def get_login_and_group(self, p):
             # so bypass it and keep default value
             return
 
-    def create_daemon(self):
+    def create_daemon(self, is_daemon=False, do_replace=False):
```
mohierf: It was to make a test ... the daemon class has this initialization parameter.  Isn't it necessary?

test/test_daemon_start.py
mohierf: Right, I will fix this

test/test_launch_daemons.py
mohierf: Ini is never needed. There are default values for every daemon.

alignak/satellite.py
mohierf: Obviously it is not enough ... I thought it was, 

It is not logical, I agree, but after some more tests, this import seems to be needed here :/

PR #546, opened 2016-11-06
Add tests for performance data

mohierf:
Fix performance data missing string strip (remaining space characters)
Add test for performance data parsing
Clean the fake_check function in AlignakTest


alignak/action.py
```diff
@@ -205,30 +205,37 @@ def get_outputs(self, out, max_plugins_output_length):
...
                 long_output.append(elts[0].strip().replace('___PROTECT_PIPE___', '|'))
                 if len(elts) > 1:
                     in_perfdata = True
                     self.perf_data += ' ' + elts[1].strip().replace('___PROTECT_PIPE___', '|')
-        # long_output is all non output and perfline, join with \n
+
+        # long_output is all non output and performance data, joined with \n
         self.long_output = '\n'.join(long_output)
+        # Get sure the performance data are stripped
+        self.perf_data = self.perf_data.strip()
```
ddurieux: it's stripped before, like in line 232 ```elts[1].strip()...``` so you strip again for nothing

test/test_perfdata_commands.py
```diff
@@ -0,0 +1,219 @@
...
+        output = """  DISK OK - free space: / 3326 MB (56%);    |   /=2643MB;5948;5958;0;5968
+/ 15272 MB (77%);
+/boot 68 MB (69%);
+/home 69357 MB (27%);
+/var/log 819 MB (84%);    | /boot=68MB;88;93;0;98
+/home=69357MB;253404;253409;0;253414
+/var/log=818MB;970;975;0;980
+        """
+        # Simulate a check executino
+        self.fake_check(svc, 0, output)
```
ddurieux: You can do this with schedule_loop function ;)

alignak/action.py
mohierf: No, if you do not add this last stripping, trailing whitespaces are not removed ... this strip is necessary.

PR #554, opened 2016-11-07
Remove unused extra_comment in downtime

ddurieux:


Seb-Solon:
This reminds me #533 but reverted. Here we remove a parameter because actually have the data within the comment object using the comment uuid. Is this attribute used in the UI / backend ? How the comment of a downtime is printed in the UI? We get the comment object?

ddurieux:
Not used, because we create downtime directly in backend so we keep information.
In fact your comment have sens with #533 but not sure we need it here

Seb-Solon:
Ok so your create both Downtime and Comment in the backend? How Alignak is aware of that then?

ddurieux:
Yep create in backend.
The arbiter backend module get all 15 secondes thw new acknowledge and downtime from backend and if have new, create external commands: https://github.com/Alignak-monitoring-contrib/alignak-module-backend/blob/develop/alignak_module_backend/arbiter/module.py#L972-L1058

PR #562, opened 2016-11-08
Improve unit tests for macro resolver

mohierf:


alignak/macroresolver.py
```diff
@@ -758,4 +779,4 @@ def _get_events_start_time():
         :rtype: int
         TODO: Implement this
         """
-        return 0
+        return 'n/a'
```
ddurieux: idem

alignak/macroresolver.py
```diff
@@ -748,7 +769,7 @@ def _get_process_start_time():
...
```
mohierf: 'n/a' is more appropriate. I will replace XxX with n/a...

PR #565, opened 2016-11-09
Fix notifications in case use external command to enable notification

ddurieux:


alignak/objects/schedulingitem.py
```diff
@@ -1791,6 +1791,12 @@ def consume_result(self, chk, notif_period, hosts,  # pylint: disable=R0915,R091
                 if enable_action:
                     self.set_myself_as_problem(hosts, services, timeperiods, bi_modulations)
 
+                # case no notification exist but notifications enables (for example, we
```
mohierf: typo: "but notifications are enabled"

PR #571, opened 2016-11-09
Closes #552: business rules tests

### LINKED ISSUE #552: Business rules feature tests
Currently, it seems to me that the business rules feature is not tested correctly (perharps not tested at all ...). We should add some tests for this feature.

### LINKED ISSUE #559: Business rules test are broken
The most simple tests with business rules does not pass. A simple Or between two services

-----

Configuration is:
```
define service{
  check_command                  check_service!ok
  host_name                      test_host_0
  service_description            db1
  use                            generic-service
}
define service{
  check_command                  check_service!ok
  host_name                      test_host_0
  service_description            db2
  use                            generic-service
}

define service{
  check_command                  bp_rule!test_host_0,db1|test_host_0,db2
  host_name                      test_host_0
  service_description            Simple_Or
  use                            generic-service
}
```

The BR associated to the service `Simple_Or` is built as a DependencyNode object that has a sons list (other DependencyNode objects). Each of its first level son has itself a Service object in its sons list.

So the BR object should contain a list of sons:
 - the first one which has a Service object  *db1* as unique son
 - the second one which has a Service object  *db2* as unique son

In the unit tests, when the configuration is parsed, this object tree is ok, but, a bit later in the unit tests, when one try to use this BR, the Service objects are no more present in the sons list ...

### LINKED ISSUE #564: Business rule output exception
When evaluating BR output, an exception is raised:
```
Error while re-evaluating business rule: 'unicode' object has no attribute 'last_hard_state_id'
```

this is because the links to host or service are now the object uuid

**Edit**: this is observed in the unit test for the BR output

mohierf:
Business rules use objects uuid
Restore and update all the business rules tests
Fix #559
Fix #564 

-----
Documented in https://github.com/Alignak-monitoring/alignak-doc/pull/14

-----

Sorry @Seb-Solon the rebase on develop included all the modifications that were merged since this morning ... the only files to review are:
 - dependencynode.py
 - macroresolver.py
 - test_business*.py

:disappointed: I do not understand why it created this mess :disappointed: and I will try to fix this!


alignak/dependencynode.py
```diff
@@ -84,7 +85,7 @@ def __init__(self, params=None):
             if 'operand' in params:
                 self.operand = params['operand']
             if 'sons' in params:
-                self.sons = [DependencyNode(elem) for elem in params['sons']]
+                self.sons = [unserialize(elem) for elem in params['sons']]
```
Seb-Solon: can you try locally without this diff ( and the next one). We the modficiation we should have only dependency node here right?

mohierf: Forget this ... we cannot even get a valid configuration :/ unserializing is necessary!

alignak/objects/schedulingitem.py
```diff
@@ -2511,6 +2519,11 @@ def get_business_rule_output(self, hosts, macromodulations, timeperiods):
         # Expands child items format string macros.
         items = self.business_rule.list_all_elements()
         for item in items:
```
mohierf: Ok

Seb-Solon: Waiting for patch.

alignak/objects/schedulingitem.py
```diff
@@ -2511,6 +2519,11 @@ def get_business_rule_output(self, hosts, macromodulations, timeperiods):
...
+            if item in hosts:
+                item = hosts[item]
```
Seb-Solon: Here you are overriding

PR #573, opened 2016-11-09
Improve tests for daemons http interfaces

mohierf:


test/test_launch_daemons.py
```diff
@@ -387,7 +393,18 @@ def _run_daemons_and_test_api(self, ssl=False):
             self.assertIsInstance(data, list, "Data is not a list!")
             self.assertEqual(set(data), expected_data, "Daemon %s has a bad API!" % name)
 
-        print("Testing get_checks on scheduler")
+        print("Testing api_full")
+        name_to_interface = {'arbiter': ArbiterInterface,
```
Seb-Solon: It think this map was useless

mohierf: You are right ... I will remove in a next PR...

PR #575, opened 2016-11-09
Add tests for host/service extinfo

mohierf:


test/cfg/extended/extended_info.cfg
```diff
@@ -0,0 +1,59 @@
...
+    host_name			host_A
+    icon_image			host.png
+    icon_image_alt	    Alt for icon.png
+    notes               Notes
+    # This parameter is already defined in the host, thus it will not overload the previous one...
+    notes_url           http://Notes_url
+    vrml_image          vrml.png
+    statusmap_image     map.png
+    2d_coords           1
+    3d_coords           2
```
ddurieux: Test with real values, like:
```
	2d_coords	100,250
	3d_coords	100.0,50.0,75.0
```

PR #576, opened 2016-11-09
Add tests for Alignak properties

mohierf:


alignak/objects/realm.py
```diff
@@ -79,7 +79,7 @@ class Realm(Itemgroup):
         'realm_name':
             StringProp(fill_brok=['full_status']),
         'alias':
-            StringProp(fill_brok=['full_status']),
+            StringProp(default=''),
```
ddurieux: Remove the fill_brok ? I think it can be interessant to the alias in broks no?

mohierf: Previously, the alias field did not even existed in the object ... it has been added fo have a certain compatibility with the other objects.

PR #580, opened 2016-11-10
Add tests for triggers and closes #586

### LINKED ISSUE #586: Triggers allow to have Python source code in objects definition
The triggers feature allow to include Python code in the object definition; such as:
```
define host{
       check_command            check_service!ok
       host_name                test_host_trigger
       use                      generic-host
       trigger                  \n\
cpu = perf(self, 'cpu')         \n\
print "Founded cpu", cpu        \n\
if cpu >= 95: \n\
\t   critical(self, 'not good! | cpu=%d' % cpu)
}
```

This code is then compiled and executed by Alignak when the trigger is used ...

----- 
**Note** that this feature is indicated as still in Beta in the Shinken doc: http://shinken.readthedocs.io/en/latest/07_advanced/triggers.html?highlight=trigger

mohierf:
Closes #586 and add tests for triggers


ddurieux:
Keep test python code in cfg, but test we can't execute it.
python code MUST NOT BE in configuration file

ddurieux:
nice ;)

PR #583, opened 2016-11-11
Add tests for escalations

mohierf:
Add tests for escalations
Fix properties test for default escalation values


alignak/objects/contact.py
```diff
@@ -224,15 +224,19 @@ def get_groupname(self):
         :return: group name
         :rtype: str
         """
-        return self.contactgroups[0]
+        if self.contactgroups:
+            return self.contactgroups[0]
+        return 'Unknown'
```
Seb-Solon: Why not '' ? For UI purpose?

alignak/objects/contact.py
```diff
@@ -224,15 +224,19 @@ def get_groupname(self):
...
     def get_groupnames(self):
         """
         Get all the groups name whose contact belongs to
         :return: comma separated list of the groups names
         :rtype: str
         """
-        return ', '.join(self.contactgroups)
+        if self.contactgroups:
+            return ', '.join(self.contactgroups)
+        return 'Unknown'
```
Seb-Solon: Why not '' ? For UI purpose?

Seb-Solon:
The rest is LGTM

alignak/objects/contact.py
mohierf: Yes for UI or macros ... make the user informed of a problem rather than hidding it.

alignak/objects/contact.py
mohierf: Yes for UI or macros ... make the user informed of a problem rather than hidding it.

PR #589, opened 2016-11-14
Add pyopenssl by default in requirements. closes #549

### LINKED ISSUE #549: Add PyOpenSSL in the project requirements
Let's assume the following:
- it is recommended to use SSL for inter-daemon communication in a distributed environment
- as stated in the requirements, we need a specific PyOpenSSL version (0.14)
- we are obliged to install and uninstall PyOpenSSL to make full tests of the application
- it is not an important constraint to have PyOpenSSL installed even if it not used

I propose to add `pyopenssl==0.14`to the requirements.txt file

@ddurieux @Seb-Solon : what about his ?

ddurieux:


Seb-Solon:
Related to alignak-monitoring/alignak-packaging#11

PR #592, opened 2016-11-14
Add tests for BI modulations

mohierf:


alignak/objects/businessimpactmodulation.py
```diff
@@ -76,7 +76,25 @@ def get_name(self):
...
+
+        * Call our parent class is_correct checker
+
+        :return: True if the configuration is correct, otherwise False
+        :rtype: bool
+        """
+
+        # Ok just put None as modulation_period, means 24x7
+        if not hasattr(self, 'modulation_period'):
+            self.modulation_period = None
```
ddurieux: It's not the best place to do this, so manage it in __init__ or perhaps if has defaut_fill or something like it (don't remember the function)

mohierf: It is done in this function for the SchedulingItem ... this is why I did here. I will move this test to hte __init__

alignak/objects/businessimpactmodulation.py
```diff
@@ -70,13 +70,22 @@ class Businessimpactmodulation(Item):
                        'modulation_period':               StringProp(default=''),
                        })
 
+    def __init__(self, params=None, parsing=True):
+        super(Businessimpactmodulation, self).__init__(params, parsing=parsing)
+
+        # Ok just put None as modulation_period, means 24x7
+        if not hasattr(self, 'modulation_period'):
+            self.modulation_period = '24x7'
```
Seb-Solon: Shouldn't this be an uuid?

mohierf: I will have a look to this but it seems that BI modulation have some specific behavior ...

PR #598, opened 2016-11-16
Clean tests and test_actions

mohierf:


test/test_commands.py
```diff
@@ -59,6 +59,76 @@ class TestCommand(AlignakTest):
...
     """
 
+    def setUp(self):
+        self.setup_with_file('cfg/cfg_commands.cfg')
+        assert self.conf_is_correct
+
+        # Our scheduler
+        self._sched = self.schedulers['scheduler-master'].sched
+
+    def test_css_in_commands(self):
```
ddurieux: Not finished test :D

alignak/action.py
```diff
@@ -339,15 +339,15 @@ def copy_shell__(self, new_i):
...
+        """Check if the command_attribute (command line) has shell characters
         Shell characters are : '!', '$', '^', '&', '*', '(', ')', '~', '[', ']',
                                '|', '{', '}', ';', '<', '>', '?', '`'
 
         :return: True if one shell character is found, False otherwise
         :rtype: bool
         """
-        for command in self.command:
-            if command in SHELLCHARS:
+        for character in self.command:
```
ddurieux: in `SHELLCHARS` we have 18 characters and in command more, so it will be more performant to do loop on `SHELLCHARS` and check if the special char is on the command

test/test_commands.py
mohierf: oups! I intended to remove this test ...

alignak/action.py
mohierf: Why not ;)

test/test_commands.py
mohierf: oh no ... the test is really finished! There is a command in the loaded configuration that includes HTML and CSS ... so the test is really present :)

ddurieux: But you have only `self.print_header()` in this test, miss the code in the test

ddurieux: Ok, but add a comment in the function description please to be enough explicit for the user read the file later ;)

PR #613, opened 2016-11-20
Clean code for potential error with strip method

mohierf:
The only goal of this PR is to manage a Landscape error.


ddurieux:
We have test run alignak with ini files, so if not works it failed :p

PR #619, opened 2016-11-21
Fix #617: remove exit_status initialization

### LINKED ISSUE #617: Hosts are always DOWN
With the most recent develop branch version, hosts are always seen as DOWN!

Thanks to @spea1

mohierf:
When actions are sent to worker and initialized in the worker, the `exit_status` must not be set with `self.exit_status = 3` because it will be set with the value contained in the params of the function. trying to initialize this field breaks the normal initialization and the action always returns an `exit_status` which is 3.

### DIFF
--- a/alignak/action.py
+++ b/alignak/action.py
@@ -271,7 +271,7 @@ class ActionBase(AlignakObject):
                      self.command, self.exit_status, self.output)
         if self.log_actions:
             logger.info("Check result for '%s': %d, %s",
-                           self.command, self.exit_status, self.output)
+                        self.command, self.exit_status, self.output)
 
     def check_finished(self, max_plugins_output_length):
         """Handle action if it is finished (get stdout, stderr, exit code...)
@@ -312,7 +312,7 @@ class ActionBase(AlignakObject):
                 self.s_time = n_child_stime - child_stime
                 if self.log_actions:
                     logger.info("Check for '%s' exited on timeout (%d s)",
-                                   self.command, self.timeout)
+                                self.command, self.timeout)
                 return
             return
 
@@ -330,7 +330,7 @@ class ActionBase(AlignakObject):
         self.exit_status = self.process.returncode
         if self.log_actions:
             logger.info("Check for '%s' exited with return code %d",
-                           self.command, self.exit_status)
+                        self.command, self.exit_status)
 
         # we should not keep the process now
         del self.process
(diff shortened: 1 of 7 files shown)


alignak/action.py
```diff
@@ -136,7 +136,6 @@ class ActionBase(AlignakObject):
     def __init__(self, params=None, parsing=True):
         super(ActionBase, self).__init__(params, parsing=parsing)
         self.creation_time = time.time()
```
ddurieux: I think if have params not define creation_time, exist_status, fill_default and log_actions, because it's filled with serialize data

alignak/dispatcher.py
```diff
@@ -80,6 +80,12 @@ def __init__(self, conf, arbiter):
         self.arbiter = arbiter
         # Pointer to the whole conf
         self.conf = conf
+        logger.warning("Dispatcher __init__: %s / %s", self.arbiter, self.conf)
+        if hasattr(self.conf, 'confs'):
+            logger.warning("Dispatch conf confs: %s", self.conf.confs)
```
Seb-Solon: warning?

alignak/objects/schedulingitem.py
```diff
@@ -1547,6 +1547,9 @@ def consume_result(self, chk, notif_period, hosts,  # pylint: disable=R0915,R091
         ok_up = self.__class__.ok_up  # OK for service, UP for host
 
         # ============ MANAGE THE CHECK ============ #
+        if 'TEST_LOG_ACTIONS' in os.environ:
+            logger.warning("Got check result: %d for '%s'",
```
mohierf: This environment variable is useful without the DEBUG logs that are too verbose. You can set TEST_LOG_ACTIONS in your shell when you need to tweak the checks commands.

alignak/dispatcher.py
mohierf: Oups! Test side-effect

alignak/action.py
mohierf: Yes, it may be better! I saw this this morning when I ran the test once more ...

alignak/objects/schedulingitem.py
Seb-Solon: Are you only using them into test? The feature of enabling debug base on file is kinda good actually but we may define a scope / use case and then implem in its own Issue. 
Anyway, my question on the debug what more because we have a warning. Is info enough or you think it's already too verbose?

PR #624, opened 2016-11-22
Process title

### LINKED ISSUE #614: Daemons process title should include the daemon name
The process title set only includes the daemon executable (eg. alignak-poller). When there are many daemons it is confusing to have several `alignak-poller` without being able to differentiate between them ... including the daemon name (eg poller-master) will make it easier.

mohierf:
Closes #614: replace process title (eg. broker) when the daemon get its configuration (eg. broker-master)

### DIFF
--- a/alignak/daemons/brokerdaemon.py
+++ b/alignak/daemons/brokerdaemon.py
@@ -460,6 +460,8 @@ class Broker(BaseSatellite):
             else:
                 name = 'Unnamed broker'
             self.name = name
+            # Set my own process title
+            self.set_proctitle(self.name)
             # local statsd
             self.statsd_host = g_conf['statsd_host']
             self.statsd_port = g_conf['statsd_port']
--- a/alignak/misc/common.py
+++ b/alignak/misc/common.py
@@ -45,9 +45,9 @@ Previously some of those variables were linked to a specific class which made no
 """
 from collections import namedtuple
 try:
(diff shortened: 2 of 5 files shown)


alignak/misc/common.py
```diff
@@ -45,9 +45,9 @@
 """
 from collections import namedtuple
 try:
```
ddurieux: It's an old code before some times ago, setproctitle not in requirements, so now we can import it without use try

alignak/daemons/brokerdaemon.py
```diff
@@ -460,6 +460,8 @@ def setup_new_conf(self):  # pylint: disable=R0915,R0912
             else:
                 name = 'Unnamed broker'
             self.name = name
+            # Set my own process title
+            self.set_proctitle(self.name)
```
ddurieux: This is managed in self.do_daemon_init_and_start() in main function. So if you add this here, it mean the new_conf modify the satellite name? but it's defined in the ini file, so I think we could not add this.

alignak/misc/common.py
mohierf: Some OSes are not compatible with the setproctitle library ... it is worth keeping this exception catch

alignak/daemons/brokerdaemon.py
mohierf: Yes but the daemon do not yet has its own name (eg. broker-master), only its type (eg. broker)

PR #633, opened 2016-11-25
Improve daemons launch test

mohierf:
More tests for daemons launch:
 - arbiter exit code (verify, bad configuration, correct configuration)
 - arbiter when other daemons are not present

### DIFF
--- a/test/test_launch_daemons.py
+++ b/test/test_launch_daemons.py
@@ -76,16 +76,7 @@ class DaemonsStartTest(AlignakTest):
             '/usr/local/var/log/alignak': '/tmp',
             '/usr/local/etc/alignak': '/tmp'
         }
-        for filename in files:
-            lines = []
-            with open(filename) as infile:
-                for line in infile:
-                    for src, target in replacements.iteritems():
-                        line = line.replace(src, target)
-                    lines.append(line)
-            with open(filename, 'w') as outfile:
-                for line in lines:
-                    outfile.write(line)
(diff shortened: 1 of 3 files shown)


test/test_launch_daemons.py
```diff
@@ -109,11 +107,79 @@ def test_arbiter_bad_configuration(self):
...
+        # copy etc config files in test/cfg/run_test_launch_daemons and change folder
+        # in the files for pid and log files
+        if os.path.exists('./cfg/run_test_launch_daemons'):
+            shutil.rmtree('./cfg/run_test_launch_daemons')
+
+        shutil.copytree('../etc', './cfg/run_test_launch_daemons')
+        files = ['cfg/run_test_launch_daemons/daemons/arbiterd.ini',
+                 'cfg/run_test_launch_daemons/arbiter/daemons/arbiter-master.cfg']
+        replacements = {
+            '/usr/local/var/run/alignak': '/tmp',
```
Seb-Solon: Bsically we prefere regenerating config over have the conf specific for the test? This will help to see a config change break but make test harder to read.

Could we factorize this in a "util" function in alignak_test or something?

mohierf: I staid with the same logic as the one firstly implemented by @ddurieux for this kind of tests, and the idea was to use the default Alignak configuration and then to adapt for the test.

Ok to make a utility function for the file replacement part.

PR #640, opened 2016-12-03
Replace grep into passwd by getent command and add quotes in default file

achaussier:


achaussier:
I think this change is not the root cause of the test failure :)

Seb-Solon:
I re-run it, turns out to be green. For your information, init script (and so alignak.in) will be drop some days. At least for from packaging as everything is systemd base. 

For the record : getent is part of glibc* packages so no new dep involved here. Looks like getent is in OpenBSD since 3.9 (pretty old i think) so no issue here too. 

LTGM. @ddurieux  any objection on the BSD part?

ddurieux:
Exist since FreeBSD 7.0, so ok for me ;)

PR #644, opened 2016-12-12
Closes #643: passive checks are immediate HARD state

### LINKED ISSUE #643: Passive checks consider max_attempts and SOFT state and they should not
When receiving some passive checks, Alignak should not consider the `max_attempts` property ... a passive check should have an immediate HARD state type except if the alignak parameter `passive_host_checks_are_soft` is set.

mohierf:
When testing max check attempts, if item is only passively checked, then raise an HARD state


mohierf:
As commented in #643, this PR is of no interest.

Issue #658, opened 2016-12-21
No pollers in the broker configuration


mohierf:
Error log in the broker when receiving its configuration:
```
[2016-12-21 18:46:51 CET] CRITICAL: [alignak.daemon] I got an unrecoverable error. I have to exit.
[2016-12-21 18:46:51 CET] CRITICAL: [alignak.daemon] You can get help at https://github.com/Alignak-monitoring/alignak
[2016-12-21 18:46:51 CET] CRITICAL: [alignak.daemon] If you think this is a bug, create a new issue including as much details as possible (version, configuration, ...
[2016-12-21 18:46:51 CET] CRITICAL: [alignak.daemon] -----
[2016-12-21 18:46:51 CET] CRITICAL: [alignak.daemon] Back trace of the error: Traceback (most recent call last):
  File "/home/alignak/alignak/alignak/daemons/brokerdaemon.py", line 886, in main
    self.setup_new_conf()
  File "/home/alignak/alignak/alignak/daemons/brokerdaemon.py", line 561, in setup_new_conf
    logger.info("[%s] pollers: %s", self.name, conf['pollers'])
KeyError: 'pollers'
```

It seems that sometimes the broker do not receive any poller configuration!

I got this with a test configuration with spare daemons

Seb-Solon:
Does that occurs only on spare or all cases?

mohierf:
AFAIR, only for spare.

PR #659, opened 2016-12-21
Test spare daemons

mohierf:
Add tests for spare daemons

@ddurieux : 

 - as of #658 this test will demonstrate the problem. Some warning logs are raised for the broker daemon because there is no pollers, receivers, ...
 - this test allows to easily stop a daemon but I do not know how to parse the log to confirm that the spare daemon started. Perharps we should add some warning logs to alert about this

### DIFF
--- a/alignak/daemons/arbiterdaemon.py
+++ b/alignak/daemons/arbiterdaemon.py
@@ -581,6 +581,7 @@ class Arbiter(Daemon):  # pylint: disable=R0902
             conf = self.new_conf
             if not conf:
                 return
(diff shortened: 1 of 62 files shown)


alignak/daemons/arbiterdaemon.py
```diff
@@ -581,6 +581,7 @@ def setup_new_conf(self):
             conf = self.new_conf
             if not conf:
                 return
+            logger.info("Sending us a configuration %s")
```
ddurieux: It's not send us, it's mode `received a configuration from arbiter master`

PR #661, opened 2016-12-21
Clean and improve some tests

mohierf:
Remove some old tests because they are already tested
Clean AlignakTest class
Improve some tests


test/alignak_test.py
```diff
@@ -18,9 +18,9 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with Alignak.  If not, see <http://www.gnu.org/licenses/>.
 
-"""
-    This file contains classes and utilities for Alignak tests modules
-"""
+#
+# This file is used to test host- and service-downtimes.
```
ddurieux: Not right, it has generic function for alignak test, not for downtime

test/test_external_commands.py
```diff
@@ -706,6 +728,17 @@ def test_change_contact_attributes(self):
         assert 65536 == contact.modified_service_attributes
 
         #  ---
+        # External command: change service custom var - undefeined variable
```
ddurieux: typo `undefeined` => `undefined`

Issue #656, opened 2016-12-21
Alignak installation directories


mohierf:
This issue is a sum-up of all the discussions regarding Alignak installation and configuration.

---
Two main problems to cope with:
 - alignak is installed from an OS distro but it must also be "installable" from a tarball/repo
 - one must be able to find out how and where alignak is installed

--- 

What I propose:
 1. a main alignak configuration file that defines where/how installation is done. 
    This file will be delivered by the OS packaging according to the file locations used by the installer.
    A default configuration file is delivered in the repo and the end-user will be able to update its content to make its own configuration if needed.

 1. alignak daemons start/stop scripts are OS dependent and must be located in the alignak packaging repository.

 1. some example scripts exist in the repo's *dev* directory. Thoses scripts can be used as examples or to start / stop alignak daemons when using the tarball/repo

I will submit a pull request according to this proposal.

PR #667, opened 2016-12-30
Update installer

mohierf:
Related with https://github.com/Alignak-monitoring/alignak-packaging/pull/26 and #656


mohierf:
@Seb-Solon : please have a look on the virtualenv test?

etc/daemons/brokerd.ini
```diff
@@ -9,7 +9,7 @@ etcdir=/usr/local/etc/alignak
 
 #-- Note that those variables:
 # 1/ are used in this file as %(workdir)s
-# 2/ are automatically updated during the Alignak installation process
+# 2/ are automatically updated during the Alignak installation process (eg. python setup.py install)
```
ddurieux: replace `python setup.py install` per `pip install .`

dev/restart_all.sh
```diff
@@ -1,6 +1,29 @@
-#!/bin/sh
+#!/bin/bash
```
ddurieux: you replace `sh` by `bash`, the unix standard is `sh`, not installed by default on Unix systems like FreeBSD

requirements.txt
```diff
@@ -4,9 +4,9 @@
 CherryPy
 requests>=2.7.0
 importlib
-pbr
```
ddurieux: If you remove `pbr` you should remove ref too in setup.py and setup.cfg

etc/daemons/brokerd.ini
mohierf: No, not necessary in this case it is the same

dev/restart_all.sh
mohierf: Yes, but as far as I remember, some features in those scripts are `bash` specific. If it is necessary to support the restricted `sh` we will log an issue

requirements.txt
mohierf: I did not yet removed in setup.py .. because some features in setup.py still require `pbr` but it is not anymore in the requirements that are used by the packaging.

etc/daemons/brokerd.ini
ddurieux: ok

dev/restart_all.sh
ddurieux: yes I will log an issue

requirements.txt
ddurieux: ah ok I see

PR #668, opened 2017-01-02
Closes #665 - manage_brok

### LINKED ISSUE #665: In broker use only manage_brok for modules loaded and not use each brok type
When daemon broker call broks in eahc module, only use only `manage_brok` and not call each type of brok

Update too the functions in example plugin

mohierf:
Clean base module
Do not manage each brok type (let the module deal with it) - closes #665

### DIFF
--- a/alignak/basemodule.py
+++ b/alignak/basemodule.py
@@ -65,11 +65,11 @@ from alignak.misc.common import setproctitle
 
 logger = logging.getLogger(__name__)  # pylint: disable=C0103
 
-# The `properties dict defines what the module can do and
+# The `properties` dict defines what the module can do and
 # if it's an external module or not.
 # pylint: disable=C0103
 properties = {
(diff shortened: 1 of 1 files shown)


alignak/basemodule.py
```diff
@@ -390,7 +408,14 @@ def _main(self):
         logger.info("Process for module %s is now running (pid=%d)", self.alias, os.getpid())
 
         # Will block here!
-        self.main()
+        try:
+            self.main()
+        except EOFError:
+            pass
+            # logger.warning('[%s] EOF exception: %s', self.alias, traceback.format_exc())
```
ddurieux: There is an exception and no message?

mohierf: Yes, it is a *usual* exception when you are reading a queue and the process is stopped ... no need to raise a log for this!

PR #731, opened 2017-02-15
clean / simplify downtimes, acnowledges and comments

ddurieux:


alignak/acknowledge.py
```diff
@@ -68,6 +69,7 @@ class Acknowledge(AlignakObject):  # pylint: disable=R0903
         'end_time': None,
         'author': None,
         'comment': None,
+        'comment_id': StringProp(default='')
```
mohierf: Why adding this parameter? And why making it a StringProp ... a string should be enough, no? Explain in the comments ...

alignak/external_command.py
```diff
@@ -1680,8 +1670,9 @@ def del_all_host_comments(self, host):
         :type host: alignak.objects.host.Host
         :return: None
         """
-        for comment in host.comments:
-            self.del_host_comment(comment)
+        comments = host.comments.keys()
```
mohierf: I do not understand... host.comments is a dictionnary, right? Indexed with the comment uuid, right? So, why not keeping the previous code that should suit for this?

alignak/external_command.py
```diff
@@ -1706,8 +1698,9 @@ def del_all_svc_comments(self, service):
         :type service: alignak.objects.service.Service
         :return: None
         """
-        for comment in service.comments:
-            self.del_svc_comment(comment)
+        comments = service.comments.keys()
+        for uuid in comments:
```
mohierf: Same comment as the previous one...

alignak/external_command.py
ddurieux: because can have the error `dictionary changed size during iteration`

alignak/acknowledge.py
ddurieux: to reference the comment added in the host/service when create the acknowledge, like this, when delete the acknowledge, we delete the comment.
For the StringProp, I added a string property like other alignak objects

mohierf: But a comment uuid is not a StringProp but a Python string ... as other alignak objects uuid

ddurieux:
I tested load retention stored in backend before this patch, works fine.
The retention seems working better with this patch, I do it many times today and no problem

Issue #422, opened 2016-10-07
Configuration with realms and poller that manage sub-realms


mohierf:
Let's assume a realm configuration with a main _All_ realm with one or more sub-realms

The poller declared in the realm _All_ can `manage_sub_realms`. As of it, it should not be necessary to declare a poller attached to the sub-realms ...

Currently, if the sub-realms do not have their own poller, the configuration is considered as invalid by the Arbiter when it is parsed.

A configuration error is raised:

```
Configuration errors:
 - Error: Hosts exist in the realm <Realm "name"=u'realm3' /> but no poller in this realm
 - Error: Hosts exist in the realm <Realm "name"=u'realm2' /> but no poller in this realm
```

ddurieux:
oh, yep it's not good ;)

mohierf:
This is why we are currently obliged to have one poller per realm ...

mohierf:
Linked to #415

PR #743, opened 2017-02-23
Fix manage realms and sub-realms when have more than 1 level. closes …

### LINKED ISSUE #415: Make a clear status on realm daemons
My current testing configuration is as follow:
- some hosts in the All realm
- some hosts in a North realm, member of All realm

With this configuration installed on a single server I am obliged to run a scheduler, a broker, a poller and a receiver for each realm.

Does it seem ok to all: @ddurieux @Seb-Solon

### LINKED ISSUE #744: Ack on sub sub realms
When the arbiter receives an external command for an host in a sub sub realm, the command is not sent to any scheduler ...

As an example, my configuration is:
- realms: All -> South -> South-East
- hosts: `pi1` and `pi2` in realm South-East
- hosts: `passive01` and `passive02` in realm South

Acknowledge sent to `passive-01` is propagated, but acknowledge for `pi1` is not propagated.

Perharps a solution with the refactoring of #743 ?

ddurieux:
…#422 closes #415 closes #744

### DIFF
--- a/alignak/objects/realm.py
+++ b/alignak/objects/realm.py
@@ -503,19 +503,4 @@ class Realms(Itemgroups):
                  realm.nb_brokers, len(realm.potential_brokers),
                  realm.nb_receivers, len(realm.potential_receivers)
                  )
-
-    def fill_potential_satellites_by_type(self, sat_type, realm, satellites):
-        """Edit potential_*sat_type* attribute to get potential satellite from upper level realms
-
-        :param sat_type: satellite type wanted
-        :type sat_type: str
-        :param realm: the realm we want to fill potential attribute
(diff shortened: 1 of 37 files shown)


alignak/objects/realm.py
```diff
@@ -364,19 +359,6 @@ class Realms(Itemgroups):
     name_property = "realm_name"  # is used for finding hostgroups
     inner_class = Realm
 
-    def get_members_by_name(self, pname):
```
mohierf: Only for cleaning I suppose ... sure it is not still used somewhere?

ddurieux: I checked in all code + test and not used

mohierf:
Tests on the demo server running correctly

PR #756, opened 2017-03-09
Host downtime acknowledges the host

### LINKED ISSUE #752: Scheduling a downtime for an host should acknowledge all its problems
As for #732 (acknowledges) we should have a similar behavior for hosts downtimed.

When scheduling a downtime for an host problem, Alignak should acknowledge the host problem and all its services that are problems. 

Currently, scheduling a downtime for an host, let the host and all its services of this host appear as problems.  A downtime scheduling is an implicit acknowledge of the probems ... if one schedule a downtime, then one is aware about the problems 😉

mohierf:
Closes #752 - scheduling a downtime for an host acknowledges the host

The acknowledge for the host is set only when the downtime starts. Note that acknowledging an host will also acknowledge all its services that are currently problems.

### DIFF
--- a/alignak/objects/schedulingitem.py
+++ b/alignak/objects/schedulingitem.py
@@ -2127,6 +2127,7 @@ class SchedulingItem(Item):  # pylint: disable=R0902
 
         }
         notif = Notification(data)
+        logger.debug("Created a %s notification: %s", self.my_type, n_type)
 
         # Keep a trace in our notifications queue
         self.notifications_in_progress[notif.uuid] = notif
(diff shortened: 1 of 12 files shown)


alignak/objects/schedulingitem.py
```diff
@@ -2724,10 +2730,19 @@ def acknowledge_problem(self, notification_period, hosts, services, sticky, noti
             self.comments[comm.uuid] = comm
             self.broks.append(self.get_update_status_brok())
             self.raise_acknowledge_log_entry()
-            return comm
         else:
-            logger.warning("Acknowledge requested for %s %s but element state is OK/UP.",
-                           self.my_type, self.get_name())
+            logger.debug("Acknowledge requested for %s %s but element state is OK/UP.",
```
ddurieux: If the host is in state OK, we don't want to acknowledge the services (lines after), I think in case we have a service in CRITICAL but the host UP, not want acknowledge them

mohierf: Yes we want ... it is much more logical! You schedule a downtime for an host and all its faulty services are then acknowledged!

PR #759, opened 2017-03-15
Improve tests (service / hostgroups)

mohierf:
Add a test to confirm that services can be inherited from an hostgroups property in the service definition

### DIFF
--- a/test/test_hostgroup.py
+++ b/test/test_hostgroup.py
@@ -265,3 +265,33 @@ class TestHostGroup(AlignakTest):
                 "test_With another Spaces"
             ) is not \
             []
+
+    def test_service_hostgroup(self):
+        """Test hosts services inherited from a hostgroups property in service definition
+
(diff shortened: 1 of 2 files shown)


test/test_hostgroup.py
```diff
@@ -265,3 +265,32 @@ def test_hostgroup_with_space(self):
...
+
+        assert len(hg.members) == 3
+        assert len(hg.hostgroup_members) == 0
+
+        assert len(hg.get_hosts()) == 3
+        print("Hostgroup hosts:")
+        for host_id in hg.members:
+            host = self.schedulers['scheduler-master'].sched.hosts[host_id]
+            print("- host: %s" % host.get_name())
+            for service_uuid in host.services:
```
ddurieux: You must add a line to test we have at least 1 service per host, otherwise we use for but if the service not exist, we can't know that

PR #785, opened 2017-05-03
Alignak configuration hook

mohierf:
Use the `get_alignak_configuration` hook function of Arbiter modules if it exists to update the global Alignak configuration parameters and macros


mohierf:
The configuration stored in the backend is correctly loaded by Alignak according to the arbiter name as a configuration identifier. When the configuration is modified by some external commands, Alignak raises an `update_program_status` brok. Unfortunately, this brok currently does not reference the arbiter name but it rather uses the scheduler name as a reference ... this is not really coherent 😞

Issue #787, opened 2017-05-03
Pylint >= 1.7


mohierf:
With the most recent version of pylint (>1.7), many errorz are raised from the source code

mohierf:
Should have been closed automatically

PR #788, opened 2017-05-03
#787: use only pylint versions < 1.7

mohierf:


ddurieux:
If there are errors, perhaps fix them no?

mohierf:
Fix them if you have some time ... I do not have currently ;) So do not yet use the newest pylint version and keep the issue opened...

PR #790, opened 2017-05-05
Improve code coverage

### LINKED ISSUE #791: Travis / Local configuration loading is different
See all the details in https://github.com/Alignak-monitoring-contrib/alignak-module-backend/issues/80

It may be interesting to have more feedback about this on other configuration than mine

@ddurieux : please confirm on your test server

mohierf:
- add some pragma no cover for features that cannot be tested in unit tests
- add some pragma no cover for simple exception catch to avoid daemons abrupt exit
- closes #791: index service templates with host/service rather than only service_description
- fix a bug in trigger functions
- fix a broken test for default properties


alignak/objects/item.py
```diff
@@ -763,11 +763,14 @@ def add_items(self, items, index_items):
         :type index_items: bool
         :return: None
         """
+        count = 0
         for i in items:
             if i.is_tpl():
                 self.add_template(i)
+                count = count + 1
```
ddurieux: You can write better: `count += 1`

alignak/trigger_functions.py
```diff
@@ -246,7 +246,9 @@ def get_custom(obj_ref, cname, default=None):
     :return:
     :rtype:
     """
+    print obj_ref
```
ddurieux: Remove this print

alignak/trigger_functions.py
```diff
@@ -246,7 +246,9 @@ def get_custom(obj_ref, cname, default=None):
...
     objs = get_objects(obj_ref)
+    print objs
```
ddurieux: Remove this print

test/test_config.py
```diff
@@ -286,16 +286,16 @@ def test_service_with_no_host(self):
...
+        # assert "Configuration in service::will_not_exist is incorrect; " \
+        #               "from: cfg/config/alignak_service_nohost.cfg:1" in \
+        #               self.configuration_errors
+        # assert "a service has been defined without host_name nor " \
+        #               "hostgroup_name, from: cfg/config/alignak_service_nohost.cfg:1" in \
+        #               self.configuration_errors
+        # assert "[service::will_not_exist] not bound to any host." in \
+        #               self.configuration_errors
+        # assert "[service::will_not_exist] no check_command" in \
+        #               self.configuration_errors
```
ddurieux: So you not test if services has errors?

test/test_config_shinken.py
```diff
@@ -0,0 +1,126 @@
...
+            print("Hosts group: %s" % item)
+        assert len(self.arbiter.conf.hostgroups) == 8
+
+        for item in self.arbiter.conf.services:
+            print("Service: %s" % item)
+        assert len(self.arbiter.conf.services) == 94
+
+        for item in self.arbiter.conf.servicegroups:
+            print("Services group: %s" % item)
+        assert len(self.arbiter.conf.servicegroups) == 5
```
ddurieux: Can you test the services properties, I think for `Cpu` for example because it be defined 2 times but with not the same template, so need to test each has the right properties.

alignak/trigger_functions.py
mohierf: Ok

alignak/trigger_functions.py
mohierf: Ok

alignak/objects/item.py
mohierf: I find it more readable and more 'pythonic' this way 😉

test/test_config.py
mohierf: Thanks to the tempalte fix, there is no more errors raised for a service that will not exist ... I remove the comments.

test/test_config_shinken.py
mohierf: This case is always tested in the test_config.py ... that a service is correctly created with its template. It is not necessary to do it again here ...

PR #797, opened 2017-05-10
Cherrypy do not log to syslog

mohierf:
Force cherrypy default log configuration: no log except for a daemon in debug mode


mohierf:
@pwgen: would your mind confirm that this removes the logs you got in your syslog?

mohierf:
Correct behavior confirmed by  @pwgen

PR #800, opened 2017-05-11
Closes #798 - detect and run missing daemons

### LINKED ISSUE #798: Missing daemons for a realm - workaround
With the whole Alignak stuff installed, it is possible to update the monitored configuration thanks to the WebUI, nice ! But ... what happens if one adds a realm in the configuration, and then an host in this new realm?

The configuration gets realoaded by the Arbiter and then its check is failing because no daemon exist for the newly created realm! All the Alignak framework is stuck and no hosts are monitored anymore !

Two ideas to be discussed:

1. before reloading a configuration, store the former configuration and restore it if the new one fails to load properly?

1. create a fake scheduler/broker/poller daemon for the newly created realm?

The first solution seems quite difficult to implement currently because of the current coding ... but the second solution would not be that difficult to implement ... to be tried

At minimum, some logs must be raised to alert about this situation !

mohierf:
For #798:
- some cleanings in the configuration parsing logs
- modifications in the daemons command line parameters (new optional port, name and log file parameters)
- configuration parsing detects missing daemons for existing realms
- the arbiter starts detected missing daemons
- the arbiter dumps the Alignak global macros defined in the configuration


etc/alignak.cfg
```diff
@@ -264,8 +264,13 @@ pack_distribution_file=/usr/local/var/lib/alignak/pack_distribution.dat
...
 # --------------------------------------------------------------------
-#
-# Those parameters are defined in the arbiterd.ini file
-# 
+# Daemons arguments
+#daemons_arguments=
+# Daemons log file
+daemons_log_folder=/usr/local/var/log/alignak
+# Default is to allocate a port number incrementally starting from the value defined here
+daemons_initial_port=7800
```
ddurieux: In the code you define per default start at 10000 and in this config you specify 7800. perhaps define the same in the 2 places by default

ddurieux:
+ there are conflict due the merge of another PR

etc/alignak.cfg
mohierf: Yes, sure 😉 I will set 7800 everywhere...

PR #809, opened 2017-05-16
A code review for less-tested classes

### LINKED ISSUE #805: Add some tests for the external commands received by the receiver daemon
Currently a large part of the receiver daemon code is not testes by the unit tests: external commands processing ... this is quite hard to test! 

Probably we should add an environment variable that will simulate the daemon receiving some external command to make the test easier?

### LINKED ISSUE #807: Add some tests for passive daemons
Currenlty, the passive daemons mode is not tested. Add some tests for passive pollers, reactionners, ...

mohierf:
Code coverage review and improvements
Closes #807: Add a test for the passive mode of the poller and reactionner daemons
Closes #805: Add tests for external commands manager (all modes)

---
**Note** that the passive daemons mode tests raised some problems in the communication between the scheduler and the daemons. I fixed the found issues and I expect that this will also clean the #817 issue


alignak/daemons/brokerdaemon.py
```diff
@@ -420,17 +423,23 @@ def get_new_broks(self, i_type='scheduler'):
                 logger.exception(exp)
                 sys.exit(1)
 
-    def get_retention_data(self):
+    def get_retention_data(self):  # pragma: no cover, useful?
         """Get all broks
 
+        TODO: using retention in the arbiter is dangerous and
```
ddurieux: It's the retention in the broker, not in the arbiter. In the broker it can make sense

mohierf: Ok, the comment is not correct. but retention in the broker does not make any sense. No more than in the arbiter because data stored are message from the queues and broks to handle ... this is very dangerous to store and reload such data!

ddurieux: Yes you;re right ;)

PR #815, opened 2017-05-16
Remove contrib folder

mohierf:
This folder contains only old stuff:
- livestatus
- zmq and TSCA clients
- scripts used to migrate from Shinken to Alignak

All this is of no interest to keep it in the current repository


ddurieux:
+1

PR #835, opened 2017-05-29
Clean alignak.cfg configuration comments

mohierf:
This PR is only for some comments modifications in the alignak.cfg file to make it more clear how to configure 😉

### DIFF
--- a/etc/alignak.cfg
+++ b/etc/alignak.cfg
@@ -10,7 +10,10 @@
 # --------------------------------------------------------------------
 
 # -------------------------------------------------------------------------
-# Monitored objects configuration part
+# Begin - Monitored objects configuration part
+# -------------------------------------------------------------------------
+# This part of the configuration file can be removed when the monitored
+# objects are stored in the Alignak backend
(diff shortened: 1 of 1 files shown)


etc/alignak.cfg
```diff
@@ -135,6 +128,25 @@ max_plugins_output_length=65536
...
+# todo: confirm the real interest of those configuration parameters!
+#interval_length=60
+#auto_reschedule_checks=1
+#auto_rescheduling_interval=1
+#auto_rescheduling_window=180
+
+# Number of interval to spread the first checks for hosts and services
+# Default is 30
+#max_service_check_spread=30
+max_service_check_spread=5
```
ddurieux: The default is 30 but we force to 5 ??? not logical :D

mohierf: This PR did not modified anything to the default configuration parameters of Alignak. This configuration exists in that form since a very long time 😉 

As such, I do not agree changing this with this PR. If you find it not logical, open an issue to update the default configuration shipped with Alignak.

PR #839, opened 2017-05-31
Migrate pylint to 1.7

mohierf:


alignak/dependencynode.py
```diff
@@ -151,20 +151,20 @@ def get_state(self, hosts, services):
             return self.get_host_node_state(host.last_hard_state_id,
                                             host.problem_has_been_acknowledged,
                                             host.in_scheduled_downtime)
-        elif self.operand == 'service':
```
ddurieux: It's not pylint want, It only say, not use: 
```
if self.operand == 'host':
    return 'yes'
else:
    return 'no'
```
but use better:
```
if self.operand == 'host':
    return 'yes'
return 'no'
```
So you can use if and elif but remove the else when have return in if or elif

mohierf: Hmm... ok. IMHO it is lass readable but I will update this

alignak/dependencynode.py
```diff
@@ -782,14 +781,14 @@ def get_host_filters(self, expr):
         flags, expr = match.groups()
         if "g" in flags:
             return [filter_host_by_group(expr)]
-        elif "r" in flags:
+        if "r" in flags:
```
ddurieux: You need re-add the `elif` in this file ;)

mohierf: Oops! I missed some...

PR #840, opened 2017-05-31
Update default test configuration for the daemons

### LINKED ISSUE #810: Test configuration with Example module
Almost all the unit tests are using a default daemon configuration defined in *tests/cfg/default/daemons* and this default configuration makes each daemon use the Example module. This is not really useful except for certain tests that need daemons to have a module.

Create two configurations:
- the default one without any modules
- the module one with modules

mohierf:
Update with the most recent alignak default configuration 

Closes #810: remove Example module declared for each daemon


mohierf:
@ddurieux : please have a look to the 2 failing tests ... I do not understand why they fail whereas they should not!

alignak/objects/config.py
```diff
@@ -1731,21 +1731,22 @@ def fill_default_realm(self):
...
-        # Check that a default realm (and only one) is defined
-        default_realms = sum(1 for realm in self.realms
-                             if hasattr(realm, 'default') and realm.default)
-        if default_realms > 1:
-            self.add_error("Error : More than one realm are set to be the default realm")
-        elif default_realms < 1:
-            self.add_error("Error : No realm is set to be the default realm")
+        # Check that a default realm (and only one) is defined and get this default realm
+        self.realms.get_default(check=True)
+        # # Check that a default realm (and only one) is defined
```
ddurieux: remove these comment lines, if not need them, delete them ;)

ddurieux (changes requested):

alignak/objects/config.py
mohierf: You are right :wink:

mohierf:
@ddurieux : imho no rebase/squash this PR. leaving the 2 commits as is is correct...

PR #841, opened 2017-05-31
Fix #821: raise notification when WARNING goes CRITICAL

### LINKED ISSUE #821: No re-notification problem when state is changing
See the test `test_3_notifications` ...

When `notification_interval` is set to 0, notifications are never re-sent to the same contacts, but when the state of the host/service changes, a notification should be raised, but it is not raised ...

What I expected was a notification for WARNING, and then one for CRITICAL, but only the first one is raised.

mohierf:
Fix #821: notify on HARD state change WARNING -> CRITICAL even if notification_interval is 0

### DIFF
--- a/alignak/objects/schedulingitem.py
+++ b/alignak/objects/schedulingitem.py
@@ -2197,13 +2197,16 @@ class SchedulingItem(Item):  # pylint: disable=R0902
 
         for contact_id in notif_contacts:
             contact = contacts[contact_id]
-            # We do not want to notify again a contact with
-            # notification interval == 0 that has been already
-            # notified. Can happen when a service exit a downtime
-            # and still in critical/warning (and not acknowledge)
+            # We do not want to notify again a contact with notification interval == 0
+            # if has been already notified except if the item hard state changed!
+            # This can happen when a service exits a downtime and it is still in
+            # critical/warning (and not acknowledge)
             if notif.type == "PROBLEM" and \
                     self.notification_interval == 0 \
+                    and self.state_type == 'HARD' and self.last_state_type == self.state_type \
+                    and self.state == self.last_state \
                     and contact.uuid in self.notified_contacts:
+                # Do not send notification
                 continue
             # Get the property name for notification commands, like
             # service_notification_commands for service
--- a/test/test_notifications.py
+++ b/test/test_notifications.py
@@ -338,37 +338,35 @@ class TestNotifications(AlignakTest):
         assert 1 == svc.current_notification_number, 'Warning HARD, must have 1 notification'
         self.assert_actions_count(1)
         self.assert_actions_match(0, 'serviceoutput WARNING', 'command')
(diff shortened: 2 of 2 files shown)


alignak/objects/schedulingitem.py
```diff
@@ -2197,13 +2197,15 @@ def scatter_notification(self, notif, contacts, notifways, timeperiods, macromod
...
-            # notification interval == 0 that has been already
-            # notified. Can happen when a service exit a downtime
-            # and still in critical/warning (and not acknowledge)
+            # We do not want to notify again a contact with notification interval == 0
+            # if has been already notified except if the item hard state changed!
+            # This can happen when a service exits a downtime and it is still in
+            # critical/warning (and not acknowledge)
             if notif.type == "PROBLEM" and \
                     self.notification_interval == 0 \
+                    and self.state_type == 'HARD' and self.state != self.last_hard_state \
```
ddurieux: I have a doubt for `elf.state != self.last_hard_state` because if you want a notification when state go fron WARNING to CRITICAL, here it validate the if and so will not send the notification

mohierf:
```
            if notif.type == "PROBLEM" and \
                    self.notification_interval == 0 \
                    and self.state_type == 'HARD' and self.state != self.last_hard_state \
                    and contact.uuid in self.notified_contacts:
                # Do not send notification
                continue
```

The problem raised in the issue #821 is fixed with this modification, there is a test that proves this and no regression in the other notifications tests.

I will have a look to this ...

ddurieux: In the issue, you say `when the state of the host/service changes, a notification should be raised` so here, if `self.state = 'CRITICAL'` and `self.last_hard_state = 'OK'` the if is verified and so it not send the notification, so it's the opposite of we want ;)

mohierf: No. It is ok, the issue is not clear enough ... it is the specific case where `notification_interval` is 0 ... that says: do not resend the notifications ... except when HARD state changes.

mohierf:
@ddurieux : would you merge this one?

alignak/objects/schedulingitem.py
ddurieux: So it must be `self.state == self.last_hard_state` not `!=`

mohierf: Hmmm.... you're fuzzing me. I will make some more tests :wink:

mohierf:
@ddurieux : I checked what this fix is intended for ...

With `notification_interval` set to 0, a contact is never re-notified. The issue in #821 is that the service goes from WARNING to CRITICAL and the contact is not notified whereas it should be because the service HARD state changed ! And then this PR fixes this problem ... so you should merge it.

mohierf:
@ddurieux: I got it ! You are right ... the test is not correct but it is because there is another problem in the item logic ...

Issue #845, opened 2017-06-01
scheduler does not start with "enable_environment_macros=1"


spea1:
scheduler does not start with "enable_environment_macros=1" in alignak.cfg

````
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler]  -> : worker zombie (now)
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler] Action: c51a0d800b4b42c5913f730b9fe662a7 (None / fork)
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler]  -> : worker zombie (now)
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler] Action: cbcfb035e56e4e2c83bed34cf4331606 (None / fork)
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler]  -> : worker zombie (now)
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler] Action: edb5eed248f248ab825837d83f6aaa45 (None / fork)
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler]  -> : worker zombie (now)
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler] Action: 749ba493e7834aa7a91a305c85cb91a9 (None / fork)
[2017-06-01 12:37:44 UTC] CRITICAL: [alignak.daemon] I got an unrecoverable error. I have to exit.
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler]  -> : worker zombie (now)
[2017-06-01 12:37:44 UTC] CRITICAL: [alignak.daemon] You can get help at https://github.com/Alignak-monitoring/alignak
[2017-06-01 12:37:44 UTC] DEBUG: [alignak.scheduler] Action: 810eaf5d0e3947f58fae921421890f74 (None / fork)
[2017-06-01 12:37:44 UTC] CRITICAL: [alignak.daemon] If you think this is a bug, create a new issue including as much detail
[2017-06-01 12:37:44 UTC] CRITICAL: [alignak.daemon] -----
[2017-06-01 12:37:44 UTC] CRITICAL: [alignak.daemon] Back trace of the error: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/alignak/daemons/schedulerdaemon.py", line 411, in main
    self.do_mainloop()
[...]

PR #848, opened 2017-06-02
[WiP] For #845, add some tests for the macros

mohierf:
This adds some tests for: 
- macros when `enable_environment_macros` is true/false
- host macros used to get the services count

--- 
Note that currently those supplementary tests do not raise the problem that @spea1 has on his configuration with the `enable_environment_macros` set to True 😯


mohierf:
@ddurieux : this PR was still WiP 😉  no matter...

Issue #851, opened 2017-06-06
Certificate file with *.csr file extension


spea1:
*.csr file extension is to be interpreted as CSR (Certificate Signing Request)

alignak/etc/daemons/*.ini
server_cert=%(etcdir)s/certs/server.**csr**

Please use for this "*.crt" or "*.cert" otherwise it will be confusing.

/usr/local/etc/alignak/certs/server.crt
-----BEGIN CERTIFICATE-----
MIIEEzCCAvugAwIBAgIBATANBgkqhkiG9w0BAQsFADCBoz

PR #853, opened 2017-06-09
Closes  #851 - default extension for certificate file (csr -> crt)

mohierf:
Change default extension for SSL certificate file. Use the more common *crt* extension rather than the confusing *csr*


mohierf:
Closes #851

PR #886, opened 2017-08-04
Stop if port not free. closes #879

### LINKED ISSUE #879: HTTP daemon port not free - stop the daemon
When a daemon raises an error because the configured HTTP port is not free, it should stop. Currently, the daemon continues its execution and waits for its configuration.

```
ERROR: [alignak.daemon] The HTTP daemon port (0.0.0.0:20000) is not free...

```

ddurieux:


ddurieux:
Miss unit tests

ddurieux:
Too complex to add a test because alignak exit before begin start logger :(

PR #899, opened 2017-08-09
Fix too many load retention and save retention. closes #898

### LINKED ISSUE #898: scheduler - new_conf - retention problems
When scheduler receive new conf, it load retention 2 times 

* first time in the `schedulerdaemon.setup_new_conf`
* second time in the `scheduler.run`

I propose remove it in `scheduler.run`

ddurieux:


alignak/daemons/schedulerdaemon.py
```diff
@@ -258,10 +258,6 @@ def setup_new_conf(self):  # pylint: disable=too-many-statements
             satellites = new_conf['satellites']
             instance_name = new_conf['instance_name']
 
-            # Ok now we can save the retention data
-            if hasattr(self.sched, 'conf'):
-                self.sched.update_retention_file(forced=True)
```
mohierf: I should not have removed this... because, when the scheduler already has a configuration and it receives a new one, this code will allow to save the retention data before loading the new configuration. Without this, we may loose some data!

PR #901, opened 2017-08-12
[WiP] - Add a multi-realms test in the load test suite

mohierf:


alignak/daemons/arbiterdaemon.py
```diff
@@ -546,16 +546,48 @@ def manage_missing_daemons(self):
...
+                    if self.my_satellites[daemon_name].stderr:
+                        for line in iter(self.my_satellites[daemon_name].stderr.readline, b''):
+                            logger.error(">>> " + line.rstrip())
                     result = False
                 else:
                     logger.info("%s running (pid=%d)",
                                 daemon_name, self.my_satellites[daemon_name].pid)
         return result
 
+    def stop_missing_daemons(self):
```
ddurieux: This function has many problems:
* in comment, it's written `stop the daemon` but in code it just look if stopped, not try stop it
* in comment we have `True if all daemons are stopped, else False`, but in code it return True only in case all daemons are running

alignak/daemons/arbiterdaemon.py
```diff
@@ -784,6 +816,9 @@ def do_loop_turn(self):
             # Main loop
             self.run()
 
+        if self.interrupted:
+            self.stop_missing_daemons()
```
ddurieux: after my previous comment, we not use the return status, perhaps manage it

mohierf:
@ddurieux, please merge this ...

ddurieux:
But you not see my comments... :p

mohierf:
I deprecate this in favor of #990

PR #903, opened 2017-08-24
Fix default value of stalking_options property

ddurieux:


ddurieux:
In the backend, the `''` value is not allowed for `stalking_options` so when import cfg we can have problem

mohierf:
The test for the default value of the properties is broken .... please fix it

PR #906, opened 2017-08-29
Fix #905: correctly manage the output encoding for external commands

### LINKED ISSUE #905: External commands passive checks encoding
When Alignak receives an external command for a passive service check that contains some special characters in the reported plugin output, the receiver daemon raises an exception when the command is submitted to the scheduler (as reported in #887).

No exception is raised when the monitoring logs are disabled for the passive checks in the `alignak.cfg` file.

This is mainly due to an incorrect management of the plugin output encoding that is of no importance if monitoring logs are not activated.

mohierf:
As reported by @spea1 ... special characters in passive plugin output may break the Alignak receiver daemon.

This pull request improves the external commands plugin output decoding and adds some tests for this problem

### DIFF
--- a/alignak/external_command.py
+++ b/alignak/external_command.py
@@ -2973,7 +2973,47 @@ class ExternalCommandManager:
         :return: None
         TODO: say that check is PASSIVE
         """
-        # raise a PASSIVE check only if needed
+        now = time.time()
+        cls = host.__class__
+
+        # If globally disabled OR host disabled, do not launch..
+        if not cls.accept_passive_checks or not host.passive_checks_enabled:
+            return
+
+        try:
+            plugin_output = plugin_output.decode('utf8', 'ignore')
+            logger.debug('%s > Passive host check plugin output: %s',
+                         host.get_full_name(), plugin_output)
+        except UnicodeError:
+            pass
+
+        # Maybe the check is just too old, if so, bail out!
+        if self.current_timestamp < host.last_chk:
+            logger.warning('%s > Passive host check is too old. Ignoring...',
+                           host.get_full_name())
+            return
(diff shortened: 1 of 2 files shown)


mohierf:
Expected behavior is confirmed by @spea1

alignak/external_command.py
```diff
@@ -2973,49 +2973,61 @@ def process_host_check_result(self, host, status_code, plugin_output):
...
+        # If globally disabled OR host disabled, do not launch..
+        if not cls.accept_passive_checks or not host.passive_checks_enabled:
+            return
+
+        try:
+            plugin_output = plugin_output.decode('utf8', 'ignore')
+            logger.debug('%s > Passive host check plugin output: %s',
+                         host.get_full_name(), plugin_output)
+        except UnicodeError:
+            pass
```
ddurieux: It can be nice to have a warning / error

alignak/external_command.py
```diff
@@ -3044,51 +3056,69 @@ def process_service_check_result(self, service, return_code, plugin_output):
...
+            logger.debug('%s > Passive service check plugin output: %s',
+                         service.get_full_name(), plugin_output)
+        except UnicodeError:
+            pass
+
+        # Maybe the check is just too old, if so, bail out!
+        if self.current_timestamp < service.last_chk:
+            logger.warning('%s > Passive service check is too old. Ignoring...',
+                           service.get_full_name())
+            return
```
ddurieux: same here

alignak/external_command.py
mohierf: It is of no interest... except for spamming the log 😉

PR #910, opened 2017-08-30
Fix #909 - old Nagios unmanaged parameters only raise a warning

### LINKED ISSUE #909: Host/service obsess Nagios parameters break the configuration check
When old Nagios parameters related to host/service obsess over feature are used in the Alignak configuration, the configuration is not considered as valid by the Alignak arbiter.

The concerned parameters are:
```
# Unmanaged parameters
use_regexp_matching=1
obsess_over_hosts=1
obsess_over_services=1
ochp_command=not_implemented
ocsp_command=not_implemented
```

Perharps that a simple Warning log would be enough for those parameters 😉

mohierf:
With recent modifications, some old Nagios parameters made the configuration being refused as invalid by the Arbiter configuration parser. a simple Warning log for each non-managed parameter is enough...

### DIFF
--- a/alignak/objects/config.py
+++ b/alignak/objects/config.py
@@ -2163,38 +2163,38 @@ class Config(Item):  # pylint: disable=R0904,R0902
         valid = True
         if self.use_regexp_matching:
             msg = "use_regexp_matching parameter is not managed."
-            logger.error(msg)
-            self.configuration_errors.append(msg)
+            logger.warning(msg)
+            self.add_warning(msg)
             valid &= False
         if getattr(self, 'obsess_over_hosts', None):
             msg = "obsess_over_hosts parameter is not managed."
-            logger.error(msg)
-            self.configuration_errors.append(msg)
+            logger.warning(msg)
+            self.add_warning(msg)
             valid &= False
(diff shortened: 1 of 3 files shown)


alignak/objects/config.py
```diff
@@ -2163,38 +2163,38 @@ def check_error_on_hard_unmanaged_parameters(self):
...
-            self.configuration_errors.append(msg)
+            logger.warning(msg)
+            self.add_warning(msg)
             valid &= False
         if getattr(self, 'ocsp_timeout', None):
             msg = "ocsp_timeout parameter is not managed."
-            logger.error(msg)
-            self.configuration_errors.append(msg)
+            logger.warning(msg)
+            self.configuration_warnings.append(msg)
```
ddurieux: couple lines before you use `self.add_warning(msg)` and here `self.configuration_warnings.append(msg)`

I could be nice to use always same code :D

mohierf: Sure it could ... but there is so many code that do not use the `add_warning` function that I prefered not to change it all. A huge *config.py* refactoring would be necessary ...

PR #916, opened 2017-09-01
Revert "[WiP] - Fix #911 - update contact notification commands properties with NW co…"

### LINKED ISSUE #911: Contact notification ways management
A `contact` object may have some notification ways and some notification commands in its configuration. the real contact notification commands are an union of the contact notification commands with those inherited from the contact notification ways.

Currently, using the contact `host_notification_commands` property only provides the contact commands and not the NW inherited commands.

This has an ugly side effect. Especially, in the `alignak-backend-import` script that relies on the object properties ...

ddurieux:
Reverts Alignak-monitoring/alignak#915

### DIFF
--- a/alignak/objects/contact.py
+++ b/alignak/objects/contact.py
@@ -315,31 +315,20 @@ class Contact(Item):
         # Oh, nobody..so NO :)
         return False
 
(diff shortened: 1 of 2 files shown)


ddurieux:
sorry it's wip :(

PR #953, opened 2017-10-30
Closes #722, set_permissions.sh script:

### LINKED ISSUE #722: dev/set_permissions.sh script with a parameter
Currently the `dev/set_permissions.sh` script:
- creates the `alignak` user
- sets the permissions for the directories located in */usr/local* ...

It may be interesting to be able to specify another root directory instead of the default */usr/local*

mohierf:
- receives 2 parameters: account name, and directory prefix
- clean user/group creation and update

This to make the script more portable ... at least the script runs correctly on CentOS 7 ;)

### DIFF
--- a/dev/set_permissions.sh
+++ b/dev/set_permissions.sh
@@ -21,74 +21,75 @@
 
 # Default is alignak account
 ACCOUNT=$1
+# Default is /usr/local account
+PREFIX=$2
 # Parse command line arguments
 if [ $# -eq 0 ]; then
     ACCOUNT="alignak"
+    PREFIX="/usr/local"
 fi
 
-## Same procedure as the one done in the debian installation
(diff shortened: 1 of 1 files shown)


dev/set_permissions.sh
```diff
@@ -21,74 +21,75 @@
 
 # Default is alignak account
 ACCOUNT=$1
+# Default is /usr/local account
+PREFIX=$2
 # Parse command line arguments
 if [ $# -eq 0 ]; then
     ACCOUNT="alignak"
+    PREFIX="/usr/local"
```
ddurieux: You will have problem if you pass only first argument and not the second

mohierf: Sure! I know ... I will add one more test in the script!

PR #956, opened 2017-11-17
Improve test run modules

mohierf:


mohierf:
Forget this... because of the daemons run refactoring

PR #957, opened 2017-11-17
Improve modules

mohierf:


alignak/modulesmanager.py
```diff
@@ -96,24 +96,24 @@ def set_daemon_name(self, daemon_name):
...
-
-    def set_max_queue_size(self, max_queue_size):
-        """Setter for max_queue_size attribute
-
-        :param max_queue_size: value to set
-        :type max_queue_size: int
-        :return: None
-        """
-        self.max_queue_size = max_queue_size
+    # def set_modules(self, modules):
```
ddurieux: Delete instead comment ;)

Issue #945, opened 2017-10-13
Daemon periodical log


mohierf:
Each daemon should raise a periodical information log; let's say every quarter ... this has two main concernes:
- information about daemon is alive in the log files
- make the python logger rotate the log files

PR #961, opened 2017-11-23
Give a try for #945: daemon periodical log

mohierf:


mohierf:
@ddurieux : I self merge this ...

PR #965, opened 2017-11-28
Closes #964 - do not raise a next_schedule brok for items that are no…

### LINKED ISSUE #964: next_check broks are not useful for passive checks
Currently, the scheduler is sending some `host_next_schedule` and `service_next_schedule` broks during its sheduling loop. Indeed, each time an host/service check is handled, this brok is raised.

For hosts / services that are not actively checked, this information is not of any interest and the broks should not be raised

mohierf:
…t actively checked


mohierf:

PR #968, opened 2017-11-28
Set module force kill delay to 60 seconds (configurable)

mohierf:


mohierf:
I merge this @ddurieux

PR #976, opened 2017-12-06
Fix erroneous log :/

mohierf:


mohierf:
I self merge this simple log fix...

PR #978, opened 2017-12-07
Improve stats

mohierf:
The envionment variable ALIGNAK_MANAGE_INTERNAL may be set to a value different of '1' to disable the Alignak internal checks logic...


mohierf:

PR #979, opened 2017-12-07
Closes #972, do not raise a configuration warning for an empty macro

### LINKED ISSUE #972: Configuration warning for empty macros
Alignak raises a configuration warning log when some macros are empty strings... no need for this ?

```
[2017-12-02 08:38:44 CET] INFO: [alignak.objects.config] Set the macro property '_DISABLED' as empty string
[2017-12-02 08:38:44 CET] INFO: [alignak.objects.config] Set the macro property '_AUTOLOGON' as empty string

```

There is no problem because a macro is an empty string ...

mohierf:


mohierf:

PR #992, opened 2018-02-02
Fix broken requirement: CherryPy must be <9.0 !

mohierf:
Restore CherryPy version < 9.0 requirement. Still necessary for some of the daemons features


mohierf:
I will forget this PR ...

PR #1106, opened 2019-05-16
Docker images

palvarezcordoba:


mohierf:
@palvarezcordoba : what is to be done with this PR? Is is still relevant if we consider #1107 ?

An sorry for the delay... :/

palvarezcordoba:
@mohierf The most important part is fixing a problem with permissions. If a run docker-compose up Alignak fails because lack of permissions on data dir.
But there is some wrong things in this PR. I'll do it better and create another PR soon.

Sorry for the inconvenience and thanks.

mohierf:
@palvarezcordoba thanks for the explanations. I understood the insterest of your endpoint.sh script +1 I worked on the Docker images tonight and I will submi some modifications probably tomorrow or Friday ... stay tuned 😉

PR #1111, opened 2019-06-13
Fix commands

mohierf:
Commands parsing and serialization did not looked very correct. This PR intends to make it better and cleaner


mohierf:
Replaced with another PR because of commits huge mess!!!

